PR #23, opened 2023-08-30
Add background job to sync dashboards to vector store

sd2k:
Very much WIP. I extracted a bunch of stuff from https://github.com/grafana/grafana/pull/69915, made use of https://github.com/grafana/grafana-plugin-examples/tree/main/examples/app-with-on-behalf-of-auth, but haven't tried running it yet.

PR #33, opened 2023-09-07
Add vector service concept, and resource endpoint to handle vector search

sd2k:
This cannibalizes some of #23 but focusses on the read path: adding a resource endpoint to the plugin to search a configured vector store using the configured embedding engine.

It adds:
- a `store` package for interacting with various vector stores
  - the interfaces are split into `ReadVectorStore` and `WriteVectorStore` so we don't need to implement everything right now (we could remove the write interface for now since it's not being used yet)
  - there's an implementation for the pending Grafana vectorapi here too; I can copy a qdrant one from a draft Grafana PR later on as another example
- an `embed` package for interacting with embedding engines
  - there's an implementation here for OpenAI embeddings (which vectorapi is also compatible with); more can be added as-and-when
- a `vector` package which wraps and combines objects from these two packages and utilises them to provide a user-friendly service to search for related text, handling the mapping from collection -> model and the embedding process
- an resource endpoint `/vector/search`, which frontend clients can use to do search for related objects in various collections
- another docker-compose service for the vectorapi to be deployed, and extra provisioning for the app including more config to talk to the vectorapi service. This will probably need to be removed for now though since vectorapi isn't OSS yet.

Some decisions made here:

- collection details are specified in config; specifically this means each collection has exactly one associated model and dimension
  - this is specified in config rather than anywhere else right now because the vector service is responsible for embedding the query using the same model and dimension as the vectors in the collection
  - in theory this could go somewhere else but we can't make many assumptions about what each vector DB implementation lets you store, so config is a fairly safe denominator
  - ideally this would be shared with whatever sync service we end up using, so that they both use the same embedding model for each collection, and the same collection names for each type of object.
- we have interfaces for `vector.Store` and `embed.Embedder` so we can have various concrete implementations, such as Qdrant/Milvus/Weaviate for the store or OpenAI/custom APIs for the embedder. This PR only adds a single one for each though, and we _may_ end up abstracting over stores in `vectorapi` instead
- there's a collection in the provisioned config named `grafana:core:dashboards`. The naming convention here is completely open to bikeshedding. The idea is to scope collections similarly to the way [Grafana Live channels are scoped](https://grafana.com/docs/grafana/latest/setup-grafana/set-up-grafana-live/#channel-structure) (the `:` can be replaced with `/` for consistency)

Testing this is a little tricky right now because we don't have open images for vectorapi, or a process for getting data _into_ the vectorapi store. One way to do so is:
- build vectorapi using `make`
- run `docker-compose up` in this repository
- run `poetry run python -m src.dashboards.index --directory ./src/dashboards/output` from the `vector-playground` directory of the `llm-experiment-lab` repository, on the `vectordb-playground-vectorapi` branch. This will send a bunch of dashboards into the `grafana:core:dashboards` collection, ready for searching
- call the resource endpoint with `curl http://localhost:3000/api/plugins/grafana-llm-app/resources/vector/search -d '{"collection": "grafana:core:dashboards", "text": "Mimir usage", "limit": 3}'`


src/plugin.json
```diff
@@ -32,5 +32,23 @@
   "dependencies": {
     "grafanaDependency": ">=9.5.2",
     "plugins": []
+  },
+  "externalServiceRegistration": {
```
yoziru, line 36: Think we can remove these permissions in this PR (only needed for sync afaik)

provisioning/plugins/grafana-llm-app.yaml
```diff
@@ -0,0 +1,21 @@
...
+      embeddings:
+        type: openai
+        openai:
+          url: http://vectorapi:8889
+      vectorStore:
+        type: grafana/vectorapi
+        grafanaVectorAPI:
+          url: http://vectorapi:8889
+        collections:
+          - name: grafana:core:dashboards
```
sd2k, line 16: When writing #43 I noticed that qdrant doesn't like a bunch of characters in collection names 🤦 so maybe we should change this?

Although they also [recommend creating _one_ collection](https://qdrant.tech/documentation/faq/qdrant-fundamentals/#how-many-collections-can-i-create) and [using payload-based partitioning](https://qdrant.tech/documentation/tutorials/multiple-partitions/) so perhaps it's not a big deal, and I should change that PR to do so...

csmarchbanks (approved):
Let's get this going and iterate on it as we get more data into vector dbs!

provisioning/plugins/grafana-llm-app.yaml
sd2k: Fixed this in 0bb2a5f, we can switch to a single collection later if we want. That _would_ let us do cool things like searching for multiple types of object matching the same search field, which would be interesting.

pkg/plugin/vector/store/store.go
```diff
@@ -0,0 +1,70 @@
...
+	Model     string `json:"model"`
+}
+
+type SearchResult struct {
+	Payload map[string]any `json:"payload"`
+	Score   float64        `json:"score"`
+}
+
+type ReadVectorStore interface {
+	Collections(ctx context.Context) ([]string, error)
```
yoziru, line 28: Do we need this method in `ReadVectorStore`? Could we just have `CollectionExists` (to check if collection exists before calculating embeddings)?

pkg/plugin/vector/store/store.go
```diff
@@ -0,0 +1,70 @@
...
+
+const (
+	VectorStoreTypeQdrant           VectorStoreType = "qdrant"
+	VectorStoreTypeGrafanaVectorAPI VectorStoreType = "grafana/vectorapi"
+)
+
+type Collection struct {
+	Name      string `json:"name"`
+	Dimension int    `json:"dimension"`
+	Model     string `json:"model"`
```
yoziru, line 19: Not sure about this particular bit, the model name isn't stored in most vector dbs. For simplicity, 
- we could go with a single model (defined at the embedder level, not at the store level)
- we could also have a separate method to get the model for a given collection
  - this could work with the single defined model at the embedder level, and then overrides based on a file for example
  - alternatively, this could call out an API for supported vector stores, that can return the model that was used to generate the embeddings (again, very few support at the collection level)

pkg/plugin/vector/store/store.go
sd2k: Switched to `CollectionExists` for simplicity

pkg/plugin/vector/store/store.go
sd2k: After chatting we decided to nest the settings a bit more and store a single model at the `vector` level, with `store` and `embed` settings nested beneath that. So it looks something like:

```yaml
jsonData:
  vector:
    model: text-embedding-ada-002
    embed:
      type: openai
    store:
      type: qdrant
      qdrant:
        address: qdrant:6334
```

yoziru (approved):
alright lgtm, provisioning needs some better docs or maybe less nesting but the core concepts work well 👍

PR #42, opened 2023-09-12
Update README Seen in the Plugin Catalog

csmarchbanks:
Mostly copy and pasting relevant sections from this repository's main README into the documentation that will be seen at https://grafana.com/grafana/plugins/grafana-llm-app/?tab=overview.

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -1,4 +1,4 @@
-# Grafana LLM app
+# Grafana LLM app (experimental)
 
 This is a Grafana application plugin which centralizes access to LLMs across Grafana.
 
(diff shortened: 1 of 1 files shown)


sd2k (commented):
LGTM! I was previously concerned that the JSX/JSON in the README would fail the plugin checks, but since https://github.com/grafana/plugin-validator/issues/131 was resolved I think we're OK.

The plugins team does try to avoid developer jargon in plugin READMEs though so we might be asked to just link to the main README instead, but let's cross that bridge when we come to it...

sd2k (approved):

PR #43, opened 2023-09-12
Add qdrant read vector store implementation

sd2k:
PR https://github.com/grafana/grafana-llm-app/pull/33 can't really be merged until we open-source and package up vectorapi, so this commit adds a qdrant implementation and modifies the provisioning file to use it (and OpenAI) by default, rather than vectorapi.

Still a little hard to test without any way of synchronising data in there, I've been working on https://github.com/grafana/grafana-llm-app/compare/vector-service-qdrant...vector-sync-2?expand=1 to make it a bit easier to get data in and test it but it's not really ideal.

### DIFF
--- a/pkg/plugin/vector/store/qdrant.go
+++ b/pkg/plugin/vector/store/qdrant.go
@@ -1,0 +1,78 @@
+package store
+
+import (
+	"context"
+
+	"github.com/grafana/grafana-plugin-sdk-go/backend/log"
+	qdrant "github.com/qdrant/go-client/qdrant"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+type qdrantSettings struct {
+	Address string
+}
+
+type qdrantStore struct {
+	conn              *grpc.ClientConn
+	collectionsClient qdrant.CollectionsClient
+	pointsClient      qdrant.PointsClient
+}
+
+func newQdrantStore(s qdrantSettings) (ReadVectorStore, func(), error) {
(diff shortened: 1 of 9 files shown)


pkg/plugin/vector/store/qdrant.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+type qdrantStore struct {
+	conn              *grpc.ClientConn
+	collectionsClient qdrant.CollectionsClient
+	pointsClient      qdrant.PointsClient
+}
+
+func newQdrantStore(s qdrantSettings) (ReadVectorStore, func(), error) {
+	conn, err := grpc.DialContext(context.Background(), s.Address, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
csmarchbanks, line 23: A TODO or issue would probably be ok for now, but we should add some auth settings. I don't love reading "insecure" in a PR :D

csmarchbanks (approved):
Looks reasonable to me and I like having a more concrete option to start with.

pkg/plugin/vector/store/qdrant.go
sd2k: Good call, I added some TLS and auth settings following [the qdrant example](https://github.com/qdrant/go-client/blob/master/examples/authentication/main.go).

sd2k:
I noticed an issue with the payload marshalling due to the payload being stored in a bunch of nested gRPC structs. Should be fixed by https://github.com/grafana/grafana-llm-app/pull/43/commits/cba49f601c39c52935b0378ea1650a77e13babfc.

PR #56, opened 2023-09-20
Added the azure openai proxy resource endpoint

edwardcqian:
Implements #55 

Added proxy implementation for Azure OpenAI

User can opt to use Azure OpenAI backend updating the following logics:
- Turn on the Use Azure OpenAI toggle
- Setting the correct URL and corresponding api-key
- Adding the model mapping (each OpenAI model should be mapped to a Azure deployment)

No changes required on the frontend.

### DIFF
--- a/pkg/plugin/resources.go
+++ b/pkg/plugin/resources.go
@@ -78,6 +78,7 @@ func newAzureOpenAIProxy() http.Handler {
 		req.URL.Scheme = "https"
 		req.URL.Host = fmt.Sprintf("%s.openai.azure.com", requestBody["resource"])
 		req.URL.Path = fmt.Sprintf("/openai/deployments/%s/chat/completions", requestBody["deployment"])
+		req.URL.RawQuery = "api-version=2023-03-15-preview"
 		req.Header.Add("api-key", settings.OpenAI.apiKey)
 
 		// Remove extra fields
--- a/pkg/plugin/resources.go
+++ b/pkg/plugin/resources.go
@@ -51,22 +51,6 @@ func (a *App) handleEcho(w http.ResponseWriter, req *http.Request) {
 }
 
 func newOpenAIProxy() http.Handler {
-	return &httputil.ReverseProxy{
-		Rewrite: func(r *httputil.ProxyRequest) {
-			config := httpadapter.PluginConfigFromContext(r.In.Context())
-			settings := loadSettings(*config.AppInstanceSettings)
-			u, _ := url.Parse(settings.OpenAI.URL)
-			r.SetURL(u)
-			r.Out.Header.Set("Authorization", "Bearer "+settings.OpenAI.apiKey)
-			organizationID := settings.OpenAI.OrganizationID
-			r.Out.Header.Set("OpenAI-Organization", organizationID)
-			r.Out.URL.Path = strings.TrimPrefix(r.In.URL.Path, "/openai")
-			log.DefaultLogger.Info("proxying to url", "url", r.Out.URL.String())
-		},
-	}
-}
-
-func newOpenAIProxy2() http.Handler {
 	director := func(req *http.Request) {
 		config := httpadapter.PluginConfigFromContext(req.Context())
 		settings := loadSettings(*config.AppInstanceSettings)
@@ -81,9 +65,13 @@ func newOpenAIProxy2() http.Handler {
 			var requestBody map[string]interface{}
 			json.Unmarshal(bodyBytes, &requestBody)
 
+			settings.OpenAI.AzureMapping = map[string]string{
+				"gpt-3.5-turbo": "gpt-35-turbo",
+			}
+
 			deployment := settings.OpenAI.AzureMapping[requestBody["model"].(string)]
 
-			req.URL.Path = fmt.Sprintf("/openai/deployments/%s/%s", deployment, strings.TrimPrefix(req.URL.Path, "/azure"))
+			req.URL.Path = fmt.Sprintf("/openai/deployments/%s/%s", deployment, strings.TrimPrefix(req.URL.Path, "/openai/v1"))
 			req.Header.Add("api-key", settings.OpenAI.apiKey)
 			req.URL.RawQuery = "api-version=2023-03-15-preview"
 
@@ -97,8 +85,8 @@ func newOpenAIProxy2() http.Handler {
 			req.URL.Path = strings.TrimPrefix(req.URL.Path, "/openai")
 			req.Header.Add("Authorization", "Bearer "+settings.OpenAI.apiKey)
 			req.Header.Add("OpenAI-Organization", settings.OpenAI.OrganizationID)
-
 		}
+		log.DefaultLogger.Info(fmt.Sprintf("UUUUUUURL: %+v", req.URL))
 	}
 	return &httputil.ReverseProxy{Director: director}
 }
--- a/pkg/plugin/resources.go
+++ b/pkg/plugin/resources.go
@@ -93,7 +93,6 @@ func newOpenAIProxy() http.Handler {
 			req.Header.Add("Authorization", "Bearer "+settings.OpenAI.apiKey)
 			req.Header.Add("OpenAI-Organization", settings.OpenAI.OrganizationID)
 		}
-		log.DefaultLogger.Info(fmt.Sprintf("UUUUUUURL: %+v", req.URL))
 	}
 	return &httputil.ReverseProxy{Director: director}
 }
--- a/pkg/plugin/resources.go
+++ b/pkg/plugin/resources.go
@@ -1,7 +1,10 @@
 package plugin
 
 import (
+	"bytes"
 	"encoding/json"
+	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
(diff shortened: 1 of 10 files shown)


sd2k (commented):
We'll need to add some options to the config page to handle this too. 

I'm not sure whether we should support multiple providers being configured at the same time... right now I don't see why not though!

pkg/plugin/settings.go
```diff
@@ -10,8 +10,10 @@ import (
 const openAIKey = "openAIKey"
 
 type OpenAISettings struct {
-	URL            string `json:"url"`
-	OrganizationID string `json:"organizationId"`
+	URL            string            `json:"url"`
+	OrganizationID string            `json:"organizationId"`
+	UseAzure       bool              `json:"azureOpenAI"`
```
sd2k, line 15: It might be nicer to use an open type here rather than a `bool`. Something like

```go
type openAIType string

const (
	openAITypeOpenAI openAIType "openai"
	openAITypeAzure  openAIType "azure"
)
```

above, then

```suggestion
	Type       openAIType            `json:"type"`
```

in the settings.

(in case in future people want support for other providers which look and act like OpenAI)

pkg/plugin/stream.go
```diff
@@ -39,29 +40,57 @@ func (a *App) SubscribeStream(ctx context.Context, req *backend.SubscribeStreamR
...
-	incomingBody := chatCompletionsRequest{Stream: true}
-	err := json.Unmarshal(req.Data, &incomingBody)
-	if err != nil {
-		return err
-	}
 
-	// Load app settings.
 	settings := loadSettings(*req.PluginContext.AppInstanceSettings)
+	var requestBody map[string]interface{}
+	json.Unmarshal(req.Data, &requestBody)
```
sd2k, line 46: We need to check the error returned here.

pkg/plugin/stream.go
```diff
@@ -39,29 +40,57 @@ func (a *App) SubscribeStream(ctx context.Context, req *backend.SubscribeStreamR
...
-	// Deserialize request data.
-	incomingBody := chatCompletionsRequest{Stream: true}
-	err := json.Unmarshal(req.Data, &incomingBody)
-	if err != nil {
-		return err
-	}
 
-	// Load app settings.
 	settings := loadSettings(*req.PluginContext.AppInstanceSettings)
+	var requestBody map[string]interface{}
```
sd2k, line 45:
```suggestion
	requestBody := map[string]interface{}{}
```

To avoid 'assignment to nil map' or anything weird like that.

pkg/plugin/resources.go
```diff
@@ -1,7 +1,10 @@
 package plugin
 
 import (
+	"bytes"
 	"encoding/json"
+	"fmt"
+	"io/ioutil"
```
sd2k, line 7: `ioutil` is deprecated, we should just use `io`.

pkg/plugin/resources.go
```diff
@@ -48,19 +51,50 @@ func (a *App) handleEcho(w http.ResponseWriter, req *http.Request) {
...
+
+		u, _ := url.Parse(settings.OpenAI.URL)
+		req.URL.Scheme = u.Scheme
+		req.URL.Host = u.Host
+
+		if settings.OpenAI.UseAzure {
+			// Map model to deployment
+			bodyBytes, _ := ioutil.ReadAll(req.Body)
+			var requestBody map[string]interface{}
+			json.Unmarshal(bodyBytes, &requestBody)
```
sd2k, line 66: Need to check the error here.

pkg/plugin/resources.go
```diff
@@ -48,19 +51,50 @@ func (a *App) handleEcho(w http.ResponseWriter, req *http.Request) {
...
+		settings := loadSettings(*config.AppInstanceSettings)
+
+		u, _ := url.Parse(settings.OpenAI.URL)
+		req.URL.Scheme = u.Scheme
+		req.URL.Host = u.Host
+
+		if settings.OpenAI.UseAzure {
+			// Map model to deployment
+			bodyBytes, _ := ioutil.ReadAll(req.Body)
+			var requestBody map[string]interface{}
```
sd2k, line 65:
```suggestion
			requestBody := map[string]interface{}{}
```

pkg/plugin/resources.go
```diff
@@ -48,19 +51,50 @@ func (a *App) handleEcho(w http.ResponseWriter, req *http.Request) {
...
+			}
+
+			req.URL.Path = fmt.Sprintf("/openai/deployments/%s/%s", deployment, strings.TrimPrefix(req.URL.Path, "/openai/v1"))
+			req.Header.Add("api-key", settings.OpenAI.apiKey)
+			req.URL.RawQuery = "api-version=2023-03-15-preview"
+
+			// Remove extra fields
+			delete(requestBody, "model")
+
+			newBodyBytes, _ := json.Marshal(requestBody)
```
sd2k, line 88: We should check the error here and log something out, at least.

pkg/plugin/stream.go
```diff
@@ -20,11 +21,11 @@ type chatCompletionsMessage struct {
...
-type chatCompletionsRequest struct {
-	Model    string                   `json:"model"`
-	Messages []chatCompletionsMessage `json:"messages"`
-	Stream   bool                     `json:"stream"`
-}
+// type chatCompletionsRequest struct {
+// 	Model    string                   `json:"model"`
+// 	Messages []chatCompletionsMessage `json:"messages"`
+// 	Stream   bool                     `json:"stream"`
+// }
```
sd2k, line 28: Yeah I think this type and the one above it can just be deleted.

pkg/plugin/stream.go
```diff
@@ -39,29 +40,57 @@ func (a *App) SubscribeStream(ctx context.Context, req *backend.SubscribeStreamR
...
+	json.Unmarshal(req.Data, &requestBody)
 
-	// Create and send OpenAI request.
-	outgoingBody, err := json.Marshal(incomingBody)
-	if err != nil {
-		return err
+	// set stream to true
+	requestBody["stream"] = true
+
+	u, _ := url.Parse(settings.OpenAI.URL)
```
sd2k, line 51: Need to handle the error.

pkg/plugin/stream.go
```diff
@@ -39,29 +40,57 @@ func (a *App) SubscribeStream(ctx context.Context, req *backend.SubscribeStreamR
...
+	} else {
+		u.Path = strings.TrimPrefix(req.Path, "openai")
 	}
-	path := strings.TrimPrefix(req.Path, "openai")
-	httpReq, err := http.NewRequestWithContext(ctx, http.MethodPost, settings.OpenAI.URL+path, bytes.NewReader(outgoingBody))
-	if err != nil {
-		return fmt.Errorf("proxy: stream: error creating request: %w", err)
+
+	outgoingBody, err = json.Marshal(requestBody)
+	httpReq, err := http.NewRequestWithContext(ctx, http.MethodPost, u.String(), bytes.NewReader(outgoingBody))
```
sd2k, line 85: Need to actually check these `err`s.

pkg/plugin/resources.go
```diff
@@ -48,19 +51,50 @@ func (a *App) handleEcho(w http.ResponseWriter, req *http.Request) {
...
+		config := httpadapter.PluginConfigFromContext(req.Context())
+		settings := loadSettings(*config.AppInstanceSettings)
+
+		u, _ := url.Parse(settings.OpenAI.URL)
+		req.URL.Scheme = u.Scheme
+		req.URL.Host = u.Host
+
+		if settings.OpenAI.UseAzure {
+			// Map model to deployment
+			bodyBytes, _ := ioutil.ReadAll(req.Body)
```
sd2k, line 64: Need to check this error

pkg/plugin/resources.go
```diff
@@ -48,19 +51,50 @@ func (a *App) handleEcho(w http.ResponseWriter, req *http.Request) {
...
-			organizationID := settings.OpenAI.OrganizationID
-			r.Out.Header.Set("OpenAI-Organization", organizationID)
-			r.Out.URL.Path = strings.TrimPrefix(r.In.URL.Path, "/openai")
-			log.DefaultLogger.Info("proxying to url", "url", r.Out.URL.String())
-		},
+	director := func(req *http.Request) {
+		config := httpadapter.PluginConfigFromContext(req.Context())
+		settings := loadSettings(*config.AppInstanceSettings)
+
+		u, _ := url.Parse(settings.OpenAI.URL)
```
sd2k, line 58: Need to check this error.

sd2k (changes requested):
This will need rebasing on `main` where there are some important changes in path handling in stream.go - basically when creating the path we should no longer use the request's path, instead we should just hardcode `/chat/completions`.

edwardcqian:
@sd2k Attempted to try handling the proxy errors, but couldn't really find a good working method. currently just collecting the errors and logging.

sd2k:
Yeah that is a bit annoying, guess there's not much more we can do? I wonder if anyone else has encountered this while using Go's HTTP proxy...

Looks like there's 1 unhandled error still left btw 🙏

edwardcqian:
> Yeah that is a bit annoying, guess there's not much more we can do? I wonder if anyone else has encountered this while using Go's HTTP proxy...
> 
> Looks like there's 1 unhandled error still left btw 🙏

Found the last unhandled error

pkg/plugin/stream.go
```diff
@@ -39,29 +29,66 @@ func (a *App) SubscribeStream(ctx context.Context, req *backend.SubscribeStreamR
...
+		}
+
+		u.Path = fmt.Sprintf("/openai/deployments/%s/chat/completion", deployment)
+		u.RawQuery = "api-version=2023-03-15-preview"
+
+		// Remove extra fields
+		delete(requestBody, "model")
+
+	} else {
+		u.Path = strings.TrimPrefix(req.Path, "openai")
```
sd2k, line 73:
```suggestion
		u.Path = "/v1/chat/completions"
```

The request path might include a UUID on it in the suffix and we can just hardcode the chat completions endpoint.

sd2k:
@edwardcqian I found a way to nicely return early if we can't parse the URL, unmarshal the body, find a matching deployment, etc. Added a few tests too. I _think_ this is good to go now, but could you test it out when you get a chance?

sd2k (dismissed):

sd2k (approved):
Let's get this merged, I'll take a look at switching the `useAzure` boolean to an open type in a separate PR.

pkg/plugin/stream.go
```diff
@@ -73,6 +73,7 @@ func (a *App) runOpenAIChatCompletionsStream(ctx context.Context, req *backend.R
 		u.Path = "/v1/chat/completions"
 	}
 
+	log.DefaultLogger.Info(fmt.Sprintf("requestBody: %s; url: %s", requestBody, u.String()))
```
sd2k, line 76: We may not want to log the request at info level, maybe debug?

PR #59, opened 2023-09-21
Add client package

sd2k:
This moves the client logic written in https://github.com/grafana/grafana-plugin-sdk-go/pull/740
to a new client subpackage in this repository.

It's based on #58 since it uses the new health check details.


pkg/client/client.go
```diff
@@ -0,0 +1,76 @@
...
+		return false, fmt.Errorf("unmarshal response: %w", err)
+	}
+	return details.OpenAIEnabled, nil
+}
+
+func (o *openAI) ChatCompletions(ctx context.Context, req openai.ChatCompletionRequest) (openai.ChatCompletionResponse, error) {
+	return o.client.CreateChatCompletion(ctx, req)
+}
+
+func (o *openAI) StreamChatCompletions(ctx context.Context, req openai.ChatCompletionRequest) (*openai.ChatCompletionStream, error) {
```
csmarchbanks, line 74: Should this be `ChatCompletionsStream` to match the return type a bit better/keep the functions together in docs?

sd2k: Mmm I named it this to match the JS library but can change it to match the go-openai library 👍

csmarchbanks (commented):
My thought on this was to move it to a sub-module inside of this repo. That way the dependency tree would be very small for anyone wanting to import it, rather than the whole plugin. What would you think of that? Pretty much just have a `llmclient/` or `grafanallm/` folder with a new go.mod in it.

sd2k:
That makes sense, I actually wanted to know whether Go would be smart enough to figure out that the subpackage didn't need all of the dependencies though. Looking at [this PR](https://github.com/grafana/grafana-plugin-sdk-go/pull/742/files) it only adds a few files to the go.sum, so maybe it's OK? Or maybe that's because all of our dependencies are already present 😂

csmarchbanks:
I think it will typically be smart enough in basic cases - until a different package also needs one of those dependencies and it needs to try to figure out which version... especially if the versions don't match or are not semver. Maybe it has gotten even smarter than that though!

sd2k:
👍 I'll move stuff around to a subdirectory/module so we don't run into any issues later.

llmclient/llmclient.go
```diff
@@ -0,0 +1,93 @@
...
+// on the given Grafana instance.
+func NewOpenAI(grafanaURL, grafanaAPIKey string) OpenAI {
+	httpClient := &http.Client{}
+	return NewOpenAIWithClient(grafanaURL, grafanaAPIKey, httpClient)
+}
+
+// NewOpenAIWithClient creates a new OpenAI client talking to the Grafana LLM app installed
+// on the given Grafana instance, using the given HTTP client.
+func NewOpenAIWithClient(grafanaURL, grafanaAPIKey string, httpClient *http.Client) OpenAI {
+	url := strings.TrimRight(grafanaURL, "/") + "/api/plugins/grafana-llm-app/resources/openai/v1"
```
yoziru, line 47: Could we put part of this path in a const? `/api/plugins/grafana-llm-app`

llmclient/llmclient.go
```diff
@@ -0,0 +1,93 @@
...
+
+type healthCheckResponse struct {
+	Details struct {
+		OpenAIEnabled bool `json:"openAI"`
+		VectorEnabled bool `json:"vector"`
+	} `json:"details"`
+}
+
+func (o *openAI) Enabled(ctx context.Context) (bool, error) {
+	req, err := http.NewRequestWithContext(ctx, "GET", o.grafanaURL+"/api/plugins/grafana-llm-app/health", nil)
```
yoziru, line 68: Same as above: Could we put part of this path in a const? `/api/plugins/grafana-llm-app`

llmclient/llmclient.go
yoziru: Don't really love the hardcoding here, but loosely held opinion so good to merge as-is 🤷

yoziru (approved):
LGTM

My main thought is that it's very openai-only based, but that applies to the current state of the plugin anyway, so this is a problem for future us type of thing (out of scope here)

llmclient/llmclient.go
sd2k: Done (including the `/resources` part) 👍

PR #60, opened 2023-09-21
Add client library to streamline using go-openai via LLM app

sd2k:
Rather than adding a whole new client struct which would basically just wrap go-openai, this adds a simple function to return an `openai.ClientConfig` which will configure the go-openai client to talk to OpenAI via our app.

It also adds a simple `OpenAIEnabled` function which uses the new plugin health check endpoint.

This is an alternative to #59


sd2k:
Closing in favour of #59 which is more restrictive. We can consider adding more functionality if needed.

PR #61, opened 2023-09-22
Set stream path to check prefix, not full string.

SandersAaronD:
Untested but seems reasonable

### DIFF
--- a/pkg/plugin/stream.go
+++ b/pkg/plugin/stream.go
@@ -32,7 +32,7 @@ func (a *App) SubscribeStream(ctx context.Context, req *backend.SubscribeStreamR
 	resp := &backend.SubscribeStreamResponse{
 		Status: backend.SubscribeStreamStatusNotFound,
 	}
-	if req.Path == openAIChatCompletionsPath {
+	if strings.HasPrefix(req.Path, openAIChatCompletionsPath) {
 		resp.Status = backend.SubscribeStreamStatusOK
 	}
 	return resp, nil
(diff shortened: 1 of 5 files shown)


pkg/plugin/stream.go
```diff
@@ -54,8 +54,8 @@ func (a *App) runOpenAIChatCompletionsStream(ctx context.Context, req *backend.R
 	if err != nil {
 		return err
 	}
-	path := strings.TrimPrefix(req.Path, "openai")
-	httpReq, err := http.NewRequestWithContext(ctx, http.MethodPost, settings.OpenAI.URL+path, bytes.NewReader(outgoingBody))
+	httpPath := strings.TrimPrefix(req.Path, "openai")
```
sd2k, line 57: This will have the UUID generated by the frontend on it now, which we don't want to include in the path we use on the request to OpenAI. I think this would actually suffice:

```suggestion
	httpPath := strings.TrimPrefix(openAIChatCompletionsPath, "openai")
```

(or some static version so we don't need to do it every request, but either works)

sd2k (changes requested):
Just 1 suggested change which makes this work locally with the both older & the fixed version of `@grafana/experimental`.

Could you also update the CHANGELOG? 🙏

SandersAaronD:
> Just 1 suggested change which makes this work locally with the both older & the fixed version of `@grafana/experimental`.
> 
> Could you also update the CHANGELOG? 🙏

This is done and version is bumped to 0.2.1, it would be helpful if we could push to release sooner rather than later.

sd2k (approved):
LGTM 👍

PR #72, opened 2023-09-25
Add release instructions to readme

SandersAaronD:
Does what it says on the tin. Probably not very useful to anyone who isn't a maintainer. If the readme gets much longer it would make sense to split it out, I think.


sd2k (approved):
Nice, thanks!

PR #73, opened 2023-09-26
Add vector fields to config page

sd2k:
Also do a bit of UI tidy up since things were getting a bit out of hand after adding Azure settings.

### DIFF
--- a/src/components/AppConfig/Vector.tsx
+++ b/src/components/AppConfig/Vector.tsx
@@ -1,0 +1,186 @@
+import React from "react";
+
+import { Field, FieldSet, Input, Select, Switch, useStyles2 } from "@grafana/ui";
+
+import { testIds } from "components/testIds";
+import { getStyles } from "./AppConfig";
+
+export interface VectorSettings {
+  // Whether the vector service should be enabled.
+  enabled?: boolean;
+  // The model used by the embedder and for embeddings stored in the store.
+  model?: string;
+  // Embedder settings.
+  embed?: EmbedderSettings;
+  // Store settings.
+  store?: StoreSettings;
+};
+
+interface OpenAIEmbedderSettings {
+  url?: string;
+};
+
+interface EmbedderSettings {
+  type?: string;
+  openai?: OpenAIEmbedderSettings;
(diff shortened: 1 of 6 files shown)


src/components/AppConfig/Vector.tsx
```diff
@@ -0,0 +1,186 @@
...
+  onChange: (settings: T) => void;
+}
+
+export function VectorConfig({ settings, onChange }: Props<VectorSettings>) {
+  return (
+    <FieldSet label="Vector Settings">
+      <Field label="Model" description="The model used by the embedder and for embeddings stored in the store">
+        <Input
+          width={60}
+          name="model"
```
yoziru, line 54: Good as-is
Maybe for the future, we could add some validation such as
- `text-embedding-ada-002` only available if using OpenAI / Azure OpenAI
- limit the models to only ones we've tried and tested (if using custom embedder) 🤔

src/components/AppConfig/Vector.tsx
```diff
@@ -0,0 +1,186 @@
...
+  // The model used by the embedder and for embeddings stored in the store.
+  model?: string;
+  // Embedder settings.
+  embed?: EmbedderSettings;
+  // Store settings.
+  store?: StoreSettings;
+};
+
+interface OpenAIEmbedderSettings {
+  url?: string;
```
yoziru, line 20: Do we need Azure OpenAI support here as well? 🤔

sd2k: Yeah we probably will do... the backend won't handle it correctly right now though, needs some unification once all this is merged.

src/components/AppConfig/Vector.tsx
sd2k: Yeah some kind of validation would be great, I think doing it via the backend health check would make most sense though since the set of models is pretty much unbounded and we don't want to be keeping our select options constantly updated.

I'm thinking that the health check would do a test chat completions run, test embedding, and test vector search, and we'd validate based on that.

yoziru (approved):

PR #76, opened 2023-09-27
Use an open 'provider' field instead of boolean Azure flag

sd2k:
This should make it easier to add more providers
in future, if required.

### DIFF
--- a/pkg/plugin/resources.go
+++ b/pkg/plugin/resources.go
@@ -211,10 +211,13 @@ func (app *App) handleVectorSearch(w http.ResponseWriter, req *http.Request) {
 
 // registerRoutes takes a *http.ServeMux and registers some HTTP handlers.
 func (a *App) registerRoutes(mux *http.ServeMux, settings Settings) {
-	if settings.OpenAI.UseAzure {
-		mux.Handle("/openai/", newAzureOpenAIProxy(settings))
-	} else {
+	switch settings.OpenAI.Provider {
+	case openAIProviderOpenAI:
 		mux.Handle("/openai/", newOpenAIProxy(settings))
+	case openAIProviderAzure:
+		mux.Handle("/openai/", newAzureOpenAIProxy(settings))
+	default:
+		log.DefaultLogger.Warn("Unknown OpenAI provider configured", "provider", settings.OpenAI.Provider)
 	}
 	mux.HandleFunc("/vector/search", a.handleVectorSearch)
 }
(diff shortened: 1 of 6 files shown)


src/components/testIds.ts
```diff
@@ -2,6 +2,7 @@ export const testIds = {
   appConfig: {
     container: 'data-testid ac-container',
     useAzureOpenAI: 'data-testid ac-use-azure-openai',
```
gerboland, line 4: remove this?

pkg/plugin/resources.go
```diff
@@ -211,10 +211,13 @@ func (app *App) handleVectorSearch(w http.ResponseWriter, req *http.Request) {
...
-	if settings.OpenAI.UseAzure {
-		mux.Handle("/openai/", newAzureOpenAIProxy(settings))
-	} else {
+	switch settings.OpenAI.Provider {
+	case openAIProviderOpenAI:
 		mux.Handle("/openai/", newOpenAIProxy(settings))
+	case openAIProviderAzure:
+		mux.Handle("/openai/", newAzureOpenAIProxy(settings))
+	default:
+		log.DefaultLogger.Warn("Unknown OpenAI provider configured", "provider", settings.OpenAI.Provider)
```
jtheory, line 220: If the provider isn't valid, this quietly logs a warning but then (I think) leaves the plugin in a broken state.

How will it fail for the developer or user who ends up here? Is there a way to make the misconfiguration more visible?

--- Oh, I see below we're forcing a provider to be chosen. So this is dead code unless there's a bug. If there's a bug, what should happen?

gerboland (approved):
lgtm!

src/components/AppConfig/OpenAI.tsx
```diff
@@ -34,13 +37,15 @@ export function OpenAIConfig({ settings, secrets, secretsSet, onChange, onChange
...
-          checked={settings.useAzure}
-          onChange={onChangeField}
+      <Field label="OpenAI Provider">
+        <Select
+          data-testid={testIds.appConfig.openAIProvider}
+          options={[
+            { label: 'OpenAI', value: 'openai' },
+            { label: 'Azure OpenAI', value: 'azure' },
+          ] as Array<SelectableValue<OpenAIProvider>>}
+          onChange={(e) => onChange({ ...settings, provider: e.value })}
```
edwardcqian, line 47:
```suggestion
          ] as Array<SelectableValue<OpenAIProvider>>}
          value={settings.provider}
          onChange={(e) => onChange({ ...settings, provider: e.value })}
```

Missing a value field to display current selected value

pkg/plugin/resources.go
sd2k: I did look for a way to handle this in the plugin but from what I can see there isn't a good way 🤔 I don't think plugin backends have the ability to validate the JSON data when it's posted to Grafana (or provided in a provisioning file); the best we could do is exit very early, but even then the logs are pretty cryptic 😞

PR #78, opened 2023-09-27
Bump version to 0.3.0

sd2k:


gerboland (approved):
Ok. No need to mention of vector settings UI?

sd2k:
Oh, hmm, guess that couldn't hurt, I'll add it in a future PR.

Issue #80, opened 2023-09-28
Add Azure OpenAI support to embedding endpoint


edwardcqian:
Currently vector embedding only support OpenAI, we may want to support Azure OpenAI as well for embeddings.

PR #85, opened 2023-10-02
Improve health check to return more granular details

sd2k:
The Dashboard AI team have requested that the plugin health check returns more details about whether the OpenAI functionality is configured, and whether that configuration is valid. This PR adds it, and similar functionality for the vector service.

It's backwards incompatible with past health checks unfortunately, but the llmclient package does include a change which should be compatible with both the current and previous version of the plugin.

We're definitely due a refactor of the way we handle OpenAI config and requests, but I think that will be part of #80.

### DIFF
--- a/pkg/plugin/health_test.go
+++ b/pkg/plugin/health_test.go
@@ -3,10 +3,31 @@ package plugin
 import (
 	"context"
 	"encoding/json"
+	"io"
+	"net/http"
+	"strings"
 	"testing"
 
+	"github.com/grafana/grafana-llm-app/pkg/plugin/vector"
+	"github.com/grafana/grafana-llm-app/pkg/plugin/vector/store"
 	"github.com/grafana/grafana-plugin-sdk-go/backend"
 )
+
+type mockHealthCheckClient struct {
+	do func(req *http.Request) (*http.Response, error)
+}
+
+func (m *mockHealthCheckClient) Do(req *http.Request) (*http.Response, error) {
+	return m.do(req)
+}
+
+type mockVectorService struct{}
+
+func (m *mockVectorService) Search(ctx context.Context, collection string, query string, topK uint64) ([]store.SearchResult, error) {
+	return []store.SearchResult{{Payload: map[string]any{"a": "b"}, Score: 1.0}}, nil
+}
+
+func (m *mockVectorService) Cancel() {}
 
 // TestCheckHealth tests CheckHealth calls, using backend.CheckHealthRequest and backend.CheckHealthResponse.
 func TestCheckHealth(t *testing.T) {
@@ -15,18 +36,23 @@ func TestCheckHealth(t *testing.T) {
 	for _, tc := range []struct {
 		name     string
 		settings backend.AppInstanceSettings
+		hcClient healthCheckClient
+		vService vector.Service
 
-		expDetails healthCheckResponse
+		expDetails healthCheckDetails
 	}{
 		{
 			name: "everything disabled",
 			settings: backend.AppInstanceSettings{
 				DecryptedSecureJSONData: map[string]string{},
 			},
-			expDetails: healthCheckResponse{
-				OpenAIEnabled: false,
-				VectorEnabled: false,
-				Version:       "unknown",
+			expDetails: healthCheckDetails{
+				OpenAI: openAIHealthDetails{
+					Error:  "No models are working",
+					Models: map[string]openAIModelHealth{},
+				},
+				Vector:  vectorHealthDetails{},
+				Version: "unknown",
 			},
 		},
 		{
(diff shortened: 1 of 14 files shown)


sd2k:
TODO:

- cache the health check result on the `App` struct, since it's expensive and slow and the app will be recreated if the plugin settings change anyway.
- Add a 'test' button to the settings UI which runs a health check (bypassing the cache?)

src/components/AppConfig/HealthCheck.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  severity = getAlertSeverity(props.status ?? 'error', props.details);
+  const message = severity === 'success' ? 'Health check succeeded!' : props.message;
+
+  return (
+    <div className="gf-form-group p-t-2">
+      <Alert severity={severity} title={message}>
+        <ShowOpenAIHealth openAI={props.details.openAI} />
+        <ShowVectorHealth vector={props.details.vector} />
+      </Alert>
```
ivanortegaalba, line 76: Thinking about the scenario where the OpenAI is correctly configured but vectorDB is ignored or misconfigured, probably it would be clearer to have two alerts, one per technology. WDYT?

ivanortegaalba, line 76: form-group can probably be replaced by a VerticalGroup. I think we're trying to deprecate those classes in favor of components

src/components/AppConfig/HealthCheck.tsx
```diff
@@ -0,0 +1,119 @@
...
+  openAI: OpenAIHealthDetails | boolean;
+  vector: VectorHealthDetails | boolean;
+  version: string;
+}
+
+interface OpenAIHealthDetails {
+  configured: boolean;
+  ok: boolean;
+  error?: string;
+  models: Record<string, OpenAIModelHealthDetails>;
```
ivanortegaalba, line 16: It could be great to have some comments here explaining the difference between ok and configured. WDYT?

src/components/AppConfig/HealthCheck.tsx
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  severity = getAlertSeverity(props.status ?? 'error', props.details);
+  const message = severity === 'success' ? 'Health check succeeded!' : props.message;
+
+  return (
+    <div className="gf-form-group p-t-2">
+      <Alert severity={severity} title={message}>
+        <ShowOpenAIHealth openAI={props.details.openAI} />
+        <ShowVectorHealth vector={props.details.vector} />
```
ivanortegaalba, line 75: Do we need to show both checks if one of them is not configured? Maybe, if it is not configured, we can just hide it, to not mislead the user.

pkg/plugin/health_test.go
```diff
@@ -87,10 +137,32 @@ func TestCheckHealth(t *testing.T) {
...
+				},
+			},
+			expDetails: healthCheckDetails{
+				OpenAI: openAIHealthDetails{
+					Configured: true,
+					OK:         true,
+					Error:      "",
+					Models: map[string]openAIModelHealth{
+						"gpt-3.5-turbo": {OK: true, Error: ""},
+						"gpt-4":         {OK: false, Error: `unexpected status code: 404: {"error": "model does not exist"}`},
```
ivanortegaalba, line 158: If we return the errors like `foo: bar: {error: baz}` it is hard to parse in the frontend. Could we just return an object? is this any type of convention?
I'd love to return to the frontend as much information as possible from OpenAI since this is part of the configuration phase.

This is the best we can do when the request fails when using the query:

src/components/AppConfig/HealthCheck.tsx
```diff
@@ -0,0 +1,119 @@
...
+          {Object.entries(openAI.models).map(([model, details], i) => (
+            <tr key={i}>
+              <td>{model}: </td>
+              <td>{details.ok ? 'OK' : `Error: ${details.error}`}</td>
+            </tr>
+          ))}
+        </tbody>
+      </table>
+    </div>
+  )
```
ivanortegaalba, line 103: Instead of using a table, what do you think about using just a bullet list with `li`? It looks like a lot of information on a corner.

Captura de pantalla 2023-10-04 a las 12 25 45

Also, I'd recommend having the banner above the button. You can miss it if you don't scroll.

src/components/AppConfig/AppConfig.tsx
```diff
@@ -86,18 +92,24 @@ export const AppConfig = ({ plugin }: AppConfigProps) => {
...
+            const result = await checkPluginHealth(plugin.meta.id);
+            setIsUpdating(false);
+            setHealthCheck(result.data);
           }}
           disabled={!updated}
         >
-          Save settings
+          Save &amp; test
         </Button>
+        {isUpdating && <Spinner />}
```
ivanortegaalba, line 109: Instead of rendering a spinner out of the button, you can use [LoadingPlaceholder](https://developers.grafana.com/ui/latest/index.html?path=/story/general-loadingplaceholder--basic) to replace the button content while loading.

src/components/AppConfig/HealthCheck.tsx
sd2k: Thanks!

src/components/AppConfig/HealthCheck.tsx
sd2k: Good point, done.

src/components/AppConfig/HealthCheck.tsx
sd2k: Good point, done! 🙂

src/components/AppConfig/HealthCheck.tsx
sd2k: Yep, makes sense and looks nicer, thanks!

src/components/AppConfig/HealthCheck.tsx
sd2k: Agreed, I've added that now.

src/components/AppConfig/AppConfig.tsx
sd2k: I've done something similar in the latest commit - take a look and see what you think?

sd2k: There might be existing UI patterns that I'm not aware of here but I haven't noticed buttons disappearing while things are loading 🤔

ivanortegaalba (approved):
✅ The experience configuring the plugin is great. The right place to have debug info.
✅ The feature still working as expected
✅ The health check works in every scenario

PR #100, opened 2023-10-16
Add support for filtered vector search

yoziru:
This PR adds support for metadata filters in vector search
Filters are based on [MongoDB's query operators](https://www.mongodb.com/docs/manual/reference/operator/query/) since some vector dbs already use this approach. For the first step, only string filtering is supported:
- $eq
- $ne
- $and
- $or

Float operators ($gte, $lte, ..) are for a later PR

Example simple

```
{
    "collection": "grafana.promql.templates",
    "query": "count over time",
    "filter": {
        "metric_type": {
            "$eq": "gauge"
        }
    }
}
```

Example OR
```json
{
    "collection": "grafana.promql.templates",
    "query": "count over time",
    "filter": {
        "$or": [
            {
                "metric_type": {
                    "$eq": "gauge"
                }
            },
            {
                "metric_type": {
                    "$eq": "histogram"
                }
            }
        ]
    }
}
```


pkg/plugin/vector/store/qdrant.go
```diff
@@ -82,14 +83,89 @@ func (q *qdrantStore) CollectionExists(ctx context.Context, collection string) (
...
+					if err != nil {
+						return nil, err
+					}
+					qdrantFilterMap.Should = append(qdrantFilterMap.Should, &qdrant.Condition{
+						ConditionOneOf: &qdrant.Condition_Filter{
+							Filter: filterMap,
+						},
+					})
+				}
+			}
```
sd2k, line 132: Do you need to handle `$and` here to match the PR description?

pkg/plugin/vector/store/qdrant.go
```diff
@@ -82,14 +83,89 @@ func (q *qdrantStore) CollectionExists(ctx context.Context, collection string) (
...
+						},
+					},
+				}
+
+				switch op {
+				case "$eq":
+					qdrantFilterMap.Must = append(qdrantFilterMap.Must, condition)
+				case "$ne":
+					qdrantFilterMap.MustNot = append(qdrantFilterMap.MustNot, condition)
+				}
```
sd2k, line 116: We should probably bail out if neither of these match, same for the `$or` / `$and` in the other case.

sd2k (commented):
A few comments inline, haven't tested this but it seems good otherwise.

sd2k (approved):
LGTM :shipit:

PR #104, opened 2023-10-17
Improve vector health check

yoziru:
Don't depend on a particular collection being there, instead check if the api/grpc connection is healthy

### DIFF
--- a/pkg/plugin/vector/service.go
+++ b/pkg/plugin/vector/service.go
@@ -13,7 +13,7 @@ import (
 
 type Service interface {
 	Search(ctx context.Context, collection string, query string, topK uint64, filter map[string]interface{}) ([]store.SearchResult, error)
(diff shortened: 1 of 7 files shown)


pkg/plugin/vector/service.go
```diff
@@ -13,6 +13,7 @@ import (
 
 type Service interface {
 	Search(ctx context.Context, collection string, query string, topK uint64, filter map[string]interface{}) ([]store.SearchResult, error)
+	Health(ctx context.Context) (bool, error)
```
sd2k, line 16: Can this just be `Health(ctx context.Context) error`? A failing health check with no error message (i.e. a `return false, nil`) isn't much use to users, we should discourage implementors from returning it

sd2k (approved):
LGTM!

sd2k:
Bleh you'll need to update cspell.config.json...

PR #106, opened 2023-10-17
docs: update provisioning docs in README

sd2k:
Commits:

- Add section on provisioning using Azure OpenAI
- Add section on provisioning vector services


jtheory (approved):
May need more context about the Qdrant example (what this is for) but let's merge this and tweak in other branch

PR #109, opened 2023-10-19
Improve config page when using Azure OpenAI

sd2k:
New screenshots:

OpenAI:
image

Azure OpenAI:
image


yoziru (approved):

jtheory:
Copying from Slack:
image

So maybe "Azure AI Service API Key" and "Azure OpenAI Language API endpoint"?
...I do notice the 3 endpoints are all the same, but I guess we just assume they might not stay that way

PR #119, opened 2023-10-23
Added auth logic for vectorapi

edwardcqian:
- added basic auth for Embedder and Store settings
- Added a Grafana Vector API Embedder option
- updated healthcheck to include Vector Embedder endpoint

## Testing instructions
1. Setup Grafana with plugin installed
2. Configure OpenAI Settings
3. Configure Grafana Vector API under Embedding settings
4. Configure Grafana Vector API under Store settings
5. Save & test, and both OpenAI and Vector checks should pass

Note: currently Azure OpenAI is not yet supported as Embedder option

### DIFF
--- a/pkg/plugin/settings.go
+++ b/pkg/plugin/settings.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 
 	"github.com/grafana/grafana-llm-app/pkg/plugin/vector"
+	"github.com/grafana/grafana-llm-app/pkg/plugin/vector/embed"
 	"github.com/grafana/grafana-plugin-sdk-go/backend"
 	"github.com/grafana/grafana-plugin-sdk-go/backend/log"
 )
@@ -46,6 +47,10 @@ func loadSettings(appSettings backend.AppInstanceSettings) Settings {
 	if settings.OpenAI.URL == "" {
 		settings.OpenAI.URL = "https://api.openai.com"
 	}
+	if settings.Vector.Embed.Type == embed.EmbedderOpenAI {
+		settings.Vector.Embed.OpenAI.URL = settings.OpenAI.URL
+		settings.Vector.Embed.OpenAI.AuthType = "openai-key-auth"
+	}
 
 	switch settings.OpenAI.Provider {
 	case openAIProviderOpenAI:
@@ -57,5 +62,6 @@ func loadSettings(appSettings backend.AppInstanceSettings) Settings {
 	}
 
 	settings.OpenAI.apiKey = appSettings.DecryptedSecureJSONData[openAIKey]
+
 	return settings
 }
--- a/pkg/plugin/vector/embed/embedder.go
+++ b/pkg/plugin/vector/embed/embedder.go
@@ -27,5 +27,7 @@ type Settings struct {
 // NewEmbedder creates a new embedder.
 func NewEmbedder(s Settings, secrets map[string]string) (Embedder, error) {
 	log.DefaultLogger.Debug("Creating OpenAI embedder")
+	// Grafana Vector API embedder is OpenAI compatible so we can reuse the client
+	// The EmbedderType is used in settings.load_settings to duplicate the correct OpenAI settings
 	return newOpenAIEmbedder(s.OpenAI, secrets), nil
 }
--- a/pkg/plugin/vector/embed/embedder.go
+++ b/pkg/plugin/vector/embed/embedder.go
@@ -9,25 +9,23 @@ import (
 type EmbedderType string
 
 const (
-	EmbedderOpenAI EmbedderType = "openai"
+	EmbedderOpenAI           EmbedderType = "openai"
+	EmbedderGrafanaVectorAPI EmbedderType = "grafana/vectorapi"
 )
 
 type Embedder interface {
 	Embed(ctx context.Context, model string, text string) ([]float32, error)
+	Health(ctx context.Context, model string) error
 }
 
 type Settings struct {
-	Type string `json:"type"`
+	Type EmbedderType `json:"type"`
 
 	OpenAI openAISettings `json:"openai"`
 }
 
 // NewEmbedder creates a new embedder.
 func NewEmbedder(s Settings, secrets map[string]string) (Embedder, error) {
-	switch EmbedderType(s.Type) {
-	case EmbedderOpenAI:
-		log.DefaultLogger.Debug("Creating OpenAI embedder")
-		return newOpenAIEmbedder(s.OpenAI, secrets), nil
-	}
-	return nil, nil
+	log.DefaultLogger.Debug("Creating OpenAI embedder")
+	return newOpenAIEmbedder(s.OpenAI, secrets), nil
 }
--- a/pkg/plugin/vector/embed/openai.go
+++ b/pkg/plugin/vector/embed/openai.go
@@ -77,7 +77,6 @@ func (o *openAIClient) Embed(ctx context.Context, model string, payload string) 
 	req.Header.Set("Content-Type", "application/json")
 	o.setAuth(req)
 
-	log.DefaultLogger.Info("MMMMMMMMMMMaking request", "req", fmt.Sprintf("%+v", req))
 	resp, err := o.client.Do(req)
 	if err != nil {
 		return nil, fmt.Errorf("make request: %w", err)
--- a/pkg/plugin/vector/embed/openai.go
+++ b/pkg/plugin/vector/embed/openai.go
@@ -13,13 +13,22 @@ import (
 )
 
 type openAISettings struct {
-	URL string `json:"url"`
+	URL           string `json:"url"`
+	AuthType      string `json:"authType"`
+	BasicAuthUser string `json:"basicAuthUser"`
+}
+
+type openAIEmbeddingsAuthSettings struct {
+	BasicAuthUser     string
+	BasicAuthPassword string
+	OpenAIKey         string
 }
 
 type openAIClient struct {
-	client *http.Client
-	url    string
-	apiKey string
+	client       *http.Client
+	url          string
+	authType     string
+	authSettings openAIEmbeddingsAuthSettings
 }
 
 type openAIEmbeddingsRequest struct {
@@ -33,6 +42,16 @@ type openAIEmbeddingsResponse struct {
 
 type openAIEmbeddingData struct {
 	Embedding []float32 `json:"embedding"`
+}
+
+func (o *openAIClient) setAuth(req *http.Request) {
+	switch o.authType {
+	case "basic-auth":
+		req.SetBasicAuth(o.authSettings.BasicAuthUser, o.authSettings.BasicAuthPassword)
+	case "openai-key-auth":
+		req.Header.Add("Authorization", "Bearer "+o.authSettings.OpenAIKey)
+	}
+
 }
 
 func (o *openAIClient) Embed(ctx context.Context, model string, payload string) ([]float32, error) {
@@ -56,9 +75,9 @@ func (o *openAIClient) Embed(ctx context.Context, model string, payload string) 
 		return nil, fmt.Errorf("create request: %w", err)
 	}
 	req.Header.Set("Content-Type", "application/json")
-	if o.apiKey != "" {
-		req.Header.Set("Authorization", "Bearer "+o.apiKey)
-	}
+	o.setAuth(req)
+
+	log.DefaultLogger.Info("MMMMMMMMMMMaking request", "req", fmt.Sprintf("%+v", req))
 	resp, err := o.client.Do(req)
 	if err != nil {
 		return nil, fmt.Errorf("make request: %w", err)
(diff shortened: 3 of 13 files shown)


jtheory:
I've not read the code, just clarifying about this:
> added basic auth support for Vector embeddings config (when basic auth is not selected it attempts to use openai api key if available)

This sounds dangerous -- are we using the OpenAI API key (set by the admin for LLM access) for vector API access as well, if the user didn't configure basic auth?

The auth we use should be exactly what the user configured, only; otherwise:
* they can't predict what's going to go where
* which makes debugging a broken setup harder
* and sometimes creates security holes (e.g.: here's a special case where sensitive auth is sent somewhere **different** than what the admin intended).

edwardcqian:
> I've not read the code, just clarifying about this:
> 
> > added basic auth support for Vector embeddings config (when basic auth is not selected it attempts to use openai api key if available)
> 
> This sounds dangerous -- are we using the OpenAI API key (set by the admin for LLM access) for vector API access as well, if the user didn't configure basic auth?
> 
> The auth we use should be exactly what the user configured, only; otherwise:
> 
> * they can't predict what's going to go where
> * which makes debugging a broken setup harder
> * and sometimes creates security holes (e.g.: here's a special case where sensitive auth is sent somewhere **different** than what the admin intended).

Good call out, this seems to be the default behaviour currently in the plugin. Make sense to add a openai option in auth, will do that for now.

We may want to revamp how the embedding logic is setup later.

csmarchbanks (commented):
It also seems like the embedder and store would always share the same basic auth credentials if set? That sounds rather limiting to me?

pkg/plugin/vector/embed/embedder.go
```diff
@@ -16,9 +16,23 @@ type Embedder interface {
...
 }
 
+type SecureJsonData struct {
+	BasicAuthPassword string `json:"basicAuthPassword"`
+	AuthToken         string `json:"authToken"`
+}
+
+type AuthSettings struct {
+	BasicAuthUser  string         `json:"basicAuthUser"`
+	SecureJsonData SecureJsonData `json:"secureJsonData"`
```
sd2k, line 26: Any secure fields definitely need to not be serialized as part of the `Settings` struct, otherwise they end up exposed to the frontend on /api/plugins/grafana-llm-app/settings:

```json
  "jsonData": {
    "vector": {
      "embed": {
        "type": "openai"
      },
      "enabled": true,
      "model": "text-embedding-ada-002",
      "store": {
        "authSettings": {
          "basicAuthUser": "aaa",
          "secureJsonData": {
            "basicAuthPassword": "secret"
          },
          "secureJsonFields": {}
        },
        "authType": "basic-auth",
        "type": "grafana/vectorapi"
      }
    }
  },
  "secureJsonFields": {
    "openAIKey": true
  }
```

Instead they should be stored in the separate secure JSON data object (the `secrets` field passed down to `newGrafanaVectorAPI`.

src/components/AppConfig/Vector.tsx
```diff
@@ -153,16 +191,19 @@ function QdrantConfig({ settings, onChange }: Props<QdrantSettings>) {
...
-        name="url"
-        data-testid={testIds.appConfig.grafanaVectorApiUrl}
-        value={settings?.url}
-        placeholder={"https://vectorapi.grafana.com"}
-        onChange={e => onChange({ ...settings, url: e.currentTarget.value })}
-      />
-    </Field>
+    <>
+      <InlineField>
+        <LegacyForms.FormField
```
sd2k, line 196: Why the switch to a legacy form field? 🤔

src/components/AppConfig/Vector.tsx
```diff
@@ -70,58 +99,67 @@ export function VectorConfig({ settings, onChange }: Props<VectorSettings>) {
...
+                    authToken: secrets.openAIKey!,
+                  },
+                  secureJsonFields: {
+                    authToken: secretsSet.openAIKey!,
+                  },
+                };
+              } else if (e.value === "VectorAPI") {
+                newSettings.authType = settings?.store?.authType;
+                newSettings.authSettings = settings?.store?.authSettings;
+              }
```
sd2k, line 155: Mentioned in another comment but e.g. here, `authSettings` shouldn't exist really, instead we should just store the inner fields on the `secrets` object. See the `openAIKey` secret field for an example!

sd2k (changes requested):

src/components/AppConfig/Vector.tsx
edwardcqian: Copied the logic from auth in datasources. Any suggestion on what to use, otherwise maybe we can drop it back to an input field.

src/components/AppConfig/AppConfig.tsx
```diff
@@ -123,7 +113,19 @@ export const getStyles = (theme: GrafanaTheme2) => ({
...
+    await updatePlugin(pluginId, data);
+
+    // Reloading the page as the changes made here wouldn't be propagated to the actual plugin otherwise.
+    // This is not ideal, however unfortunately currently there is no supported way for updating the plugin state.
+    window.location.reload();
+  } catch (e) {
+    console.error('Error while updating the plugin', e);
+  }
+};
+
```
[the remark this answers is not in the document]
edwardcqian: I think this might be accidental

src/components/AppConfig/Vector.tsx
sd2k: Maybe just an input field tbh 🤷

pkg/plugin/vector/service.go
```diff
@@ -50,6 +58,8 @@ func NewService(s VectorSettings, secrets map[string]string) (Service, error) {
 		log.DefaultLogger.Warn("No vector store configured")
 		return nil, nil
 	}
+
+	log.DefaultLogger.Info("Vector service created", "model", fmt.Sprintf("EEEEEEEE%+v\nSSSSSSS%+v", em, st))
```
sd2k, line 62: note: remove this before merge

pkg/plugin/vector/store/vectorapi.go
```diff
@@ -97,9 +131,15 @@ func (g *grafanaVectorAPI) Health(ctx context.Context) error {
...
-		client: &http.Client{},
-		url:    s.URL,
-	}
+		client:   &http.Client{},
+		url:      s.URL,
+		authType: authType,
+		authSettings: grafanaVectorAPIAuthSettings{
+			BasicAuthUser:     authS.BasicAuthUser,
+			BasicAuthPassword: secrets["basicAuthPassword"],
+			AuthHeaderValue:   secrets["authHeaderValue"],
```
sd2k, line 142: We should factor these secret keys out into constants and scope them (e.g. `grafanaVectorAPIBasicAuthPassword`?) since secrets are global and we may have some other section that needs to add a basic auth password later.

src/components/AppConfig/AppConfig.tsx
```diff
@@ -69,45 +69,59 @@ export const AppConfig = ({ plugin }: AppConfigProps) => {
...
-            setHealthCheck(undefined);
+          onClick={() => {
             let key: keyof SecretsSet;
             const secureJsonData: Secrets = {};
             for (key in configuredSecrets) {
               // Only include secrets that are not already configured on the backend,
               // otherwise we'll overwrite them.
               if (!configuredSecrets[key]) {
                 secureJsonData[key] = newSecrets[key];
+                console.log("Setting secret", key, newSecrets[key]);
```
sd2k, line 111: note: remove before merge

src/components/AppConfig/AppConfig.tsx
```diff
@@ -69,45 +69,59 @@ export const AppConfig = ({ plugin }: AppConfigProps) => {
...
               }
             }
-            await updatePlugin(plugin.meta.id, {
+            console.log("Saving settings", settings, secureJsonData);
```
sd2k, line 114: note: remove before merge

src/components/AppConfig/Vector.tsx
```diff
@@ -1,29 +1,31 @@
...
+import { TokenAuthConfig } from "./AuthSettings/TokenAuth";
 
 export interface VectorSettings {
   // Whether the vector service should be enabled.
   enabled?: boolean;
   // The model used by the embedder and for embeddings stored in the store.
   model?: string;
   // Embedder settings.
-  embed?: EmbedderSettings;
+  embedType?: string;
```
sd2k, line 17: Hmm I'd be tempted to keep this as the `EmbedderSettings` object with just the `type` key, in case we want to add more to it in future (e.g. supporting more than OpenAI and storing some data for the new embedder type).

src/components/AppConfig/Vector.tsx
```diff
@@ -36,92 +38,93 @@ interface GrafanaVectorAPISettings {
...
+  onChange: (settings: VectorSettings) => void;
+  onChangeSecrets: (secrets: Secrets) => void;
+}) {
   return (
     <FieldSet label="Vector Settings">
 
-      <Field label="Enabled" description="Enable vector database powered features.">
-        <Checkbox
+      <InlineField label="Enabled" tooltip="Enable vector database powered features.">
+        <InlineSwitch
```
sd2k, line 63: This was intentionally moved to a `Checkbox` rather than an `InlineSwitch` in a [previous commit](2d0d8804) after a review comment from a frontend dev mentioned that it's the recommended UX in Grafana.

In general it's easier to review if these other UI changes aren't scattered throughout the PR 😢

gerboland (commented):
I've some tweaks suggested inline. Testing, this appears to work as desired, with some rough edges that can be improved later.

gerboland:
Additional thought for later PR: if we're support Qdrant locally, it's a bit odd not to support their cloud offering. May need the auth stuff for that too.

src/components/AppConfig/Vector.tsx
```diff
@@ -36,92 +41,93 @@ interface GrafanaVectorAPISettings {
...
+        
+            onChange({...settings?.vector, embed: newSettings});
+          }}
         />
       </Field>
-      {settings?.type === "openai" && (
-        <OpenAIEmbedderConfig
-          settings={settings.openai}
-          onChange={openai => onChange({ ...settings, openai })}
+      <Field label="Embedding Model" description="The model used by the embedding API and for embeddings stored in the store">
```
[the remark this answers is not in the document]
edwardcqian: Agreed, exposing model in the settings isn't the right design.

There shouldn't be a reason for the user to specify the model at all. The model used should be tied to the collection being queried.

src/components/AppConfig/AuthSettings/TokenAuth.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
```
yoziru, line 1: Just to keep this PR smaller, could we get rid of token auth entirely since it's not needed atm?

pkg/plugin/vector/embed/embedder.go
```diff
@@ -2,32 +2,36 @@ package embed
...
+	EmbedderVectorAPI EmbedderType = "grafana-vector-api"
 )
 
-type Embedder interface {
-	Embed(ctx context.Context, model string, text string) ([]float32, error)
-}
+type EmbedderAuthType string
+
+const (
+	EmbedderAuthTypeBasicAuth EmbedderAuthType = "basic-auth"
```
yoziru, line 17: These types feel like generic types e.g. `AuthTypeBasicAuth ` instead of `EmbedderAuthTypeBasicAuth`
Maybe move to settings.go?

pkg/plugin/settings.go
```diff
@@ -48,14 +50,38 @@ func loadSettings(appSettings backend.AppInstanceSettings) Settings {
...
 
 	settings.OpenAI.apiKey = appSettings.DecryptedSecureJSONData[openAIKey]
-	return settings
+
+	// embedder settings
+	// If the embedding setup option is set to OpenAI, then we need to use openAIKey and URL
+	// Otherwise we use the auth setting and URL from the vector store.
+	var authType embed.EmbedderAuthType
+	var URL string
+	embedderType := embed.EmbedderType(settings.Vector.Embed.Type)
```
yoziru, line 68: Could we use the same case with default fallback here to avoid weird errors? (e.g. fallback to OpenAI if not provided)

pkg/plugin/vector/embed/embedder.go
```diff
@@ -9,25 +9,23 @@ import (
...
 // NewEmbedder creates a new embedder.
 func NewEmbedder(s Settings, secrets map[string]string) (Embedder, error) {
-	switch EmbedderType(s.Type) {
-	case EmbedderOpenAI:
-		log.DefaultLogger.Debug("Creating OpenAI embedder")
-		return newOpenAIEmbedder(s.OpenAI, secrets), nil
-	}
-	return nil, nil
+	log.DefaultLogger.Debug("Creating OpenAI embedder")
+	return newOpenAIEmbedder(s.OpenAI, secrets), nil
```
csmarchbanks, line 30: A comment here saying that vector api looks like OpenAI so we can reuse the client might be nice.

csmarchbanks (approved):
One small comment but this seems like a reasonable start.

pkg/plugin/vector/embed/openai.go
```diff
@@ -86,12 +104,22 @@ func (o *openAIClient) Embed(ctx context.Context, model string, payload string)
 	return body.Data[0].Embedding, nil
 }
 
+func (o *openAIClient) Health(ctx context.Context, model string) error {
+	_, err := o.Embed(ctx, model, "Hello, world!")
```
yoziru, line 108: This part is still failing, when I log it it seems the model passed here is empty when hitting Save & test on a fresh install.
```
Vector service health check failed.
Error: vector service health check failed: embedder health: got non-2xx status from OpenAI
```

This seems like a front-end issue though. If I change the model used (e.g. `BAAI/bge-base-en-v1.5`) and change it back to the original (e.g. `BAAI/bge-small-en-v1.5`) it does work

src/components/AppConfig/Vector.tsx
```diff
@@ -30,97 +40,140 @@ interface QdrantSettings {
...
 
-export function VectorConfig({ settings, onChange }: Props<VectorSettings>) {
+export function VectorConfig({ settings, secrets, secretsSet, onChange, onChangeSecrets }: {
+  settings?: AppPluginSettings;
+  secrets: Secrets;
+  secretsSet: SecretsSet;
+  onChange: (settings: VectorSettings) => void;
+  onChangeSecrets: (secrets: Secrets) => void;
+}) {
+  const modelValues = settings?.vector?.embed?.type === "openai" ? "text-embedding-ada-002" : settings?.vector?.embed?.type === "grafana/vectorapi" ? (settings?.vector?.model ?? "BAAI/bge-small-en-v1.5"): ""
```
yoziru, line 61: Somewhere here things are not working (see comment above)

yoziru (changes requested):

yoziru (approved):
🚀

PR #131, opened 2023-10-30
Fix health API endpoint

annanay25:
Correcting API endpoint for the `/health` call.

### DIFF
--- a/llmclient/llmclient.go
+++ b/llmclient/llmclient.go
@@ -15,6 +15,7 @@ import (
 )
 
 const (
+	appPrefix          = "/api/plugins/grafana-llm-app"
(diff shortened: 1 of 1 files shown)


llmclient/llmclient.go
```diff
@@ -15,6 +15,7 @@ import (
 )
 
 const (
+	appPrefix          = "/api/plugins/grafana-llm-app"
 	appResourcesPrefix = "/api/plugins/grafana-llm-app/resources"
```
csmarchbanks, line 19: Should we use `appPrefix` to build this up?

csmarchbanks (approved):
One small thought but fine either way 👍

PR #132, opened 2023-10-31
Merge in frontend components from experimental and the llm examples app

SandersAaronD:
Merging in these repositories allows us to do end to end testing in one place and modify builds without worrying that we're breaking other components.


csmarchbanks (commented):
I am ok with moving the examples into this repository, but I am not sure about the experimental code. In our initial discussions with the Grafana team there was a consensus that we wanted the frontend features in the experimental repository, not as a separate npm package that needed importing.

We should still be able to do end to end testing by depending on the code in experimental correct? Any change in this repo should be backwards compatible anyway, so testing with a slightly older version is possibly for the best.

PR #142, opened 2023-11-08
Only cache healthy healthcheck results

csmarchbanks:
Right now we cache results even if they come back unhealthy. This means that if OpenAI is down when a grafana instance restarts and is first queried the llm plugin will not be useful for the lifecycle of the instance. Instead, only cache if the results come back healthy.

It also breaks the cache into two parts so we can separately cache for a healthy OpenAI configuration vs VectorAPI configuration.


sd2k (approved):
LGTM

PR #151, opened 2023-11-14
Update provision README and files for vector store/embedder

edwardcqian:
Added examples of how to setup with `openai+qdrant` and `grafana/vectorapi` for vector provisioning

Removed unneeded provisioning for openai vector embedder (uses openai completion settings by default)

updated docker-compose and provisioning local dev to use grafana/vectorapi by default.


bohandley (approved):
with the following provision file and running qdrant on port 6334 for grpc it works! but the template needs to be set up so this work is good but the PromQL setup will need another issue fix.

```
apiVersion: 1

# run qdrant docker with 
# docker run -p 6334:6334 qdrant/qdrant
apps:
  - type: grafana-llm-app
    jsonData:
      openAI:
        provider: openai
        url: https://api.openai.com
        organizationId: $OPENAI_ORG
      vector:
        enabled: true
        model: text-embedding-ada-002
        embed:
          type: openai
        store:
          type: qdrant
          qdrant:
            address: localhost:6334

    secureJsonData:
      openAIKey: $OPENAI_KEY
```

jtheory:
We want to help OSS users (or internal users testing locally) to get the PromQL query advisor running -- that'll be really hard with Qdrant! I've updated the readme to guide people better. More iterations needed still.

ioanarm:
@edwardcqian I am merging this for now, so we can have a more up to date README for the day. We will iterate on the steps within the day.

PR #152, opened 2023-11-14
Changed package manager

edwardcqian:
Drone complaining about package manager during intialization 

```
+ yarn install --frozen-lockfile --no-progress
error This project's package.json defines "packageManager": "yarn@npm@8.19.3". However the current global version of Yarn is 1.22.20.
```

Changed package manger to yarn

Note: Seems CI always ran yarn and only recently broke.


csmarchbanks (commented):
Wouldn't we at least need a yarn.lock in here to be able to use yarn and expect reproducible builds?

edwardcqian:
Initializes with the specified yarn version with yarn.lock is missing. initializes with `1.22.21` when yarn.lock is present, which is the version that seems to break things

sd2k:
There are a few references to using `npm` left, at least:

- all the commands in README.md
- package-lock.json

edwardcqian:
Urgh, Grafana API compatibility check defaults to `npm`

sd2k:
That should be fixable at least, the compat check is a GitHub action so is [here](https://github.com/grafana/grafana-llm-app/blob/fcd9f2169215ed176b61910ab6a52e7f95f2004b/.github/workflows/is-compatible.yml#L14-L19) rather than outside of our control.

SandersAaronD:
This is currently blocked on this: https://github.com/grafana/grafana-drone-extension/issues/451

Would be able to do a number of helpful things if we got the package build actually updated so that it didn't error so easily, basically the build system is the current problem.

SandersAaronD:
Closing this, was intended to accomplish what is currently being done in #235

PR #205, opened 2024-01-29
feat: add Qdrant API key field to qdrant settings

### LINKED ISSUE #200: Vector settings in app do not allow for qdrantApiKey for Qdrant
"If you're using a secure connection you can set the qdrantApiKey field in secureJsonData to provide an API key with each request."

This option does not seem to display in the plug-in configuration.

Select Qdrant, paste the address, and select secure does not offer a place to put an API key.

sd2k:
The Qdrant settings interface did not allow users to modify the qdrantApiKey field, even though this field is understood by the plugin backend and documented in the README. This PR adds the API key field to the Qdrant config when the 'secure' checkbox is checked.

Fixes #200.

### DIFF
--- a/src/components/AppConfig/AppConfig.tsx
+++ b/src/components/AppConfig/AppConfig.tsx
@@ -18,6 +18,7 @@ export interface AppPluginSettings {
 
 export type Secrets = {
   openAIKey?: string;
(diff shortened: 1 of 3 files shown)


yoziru (approved):
LGTM

PR #209, opened 2024-02-02
Push state to GCom

gerboland:
To push state to GCom, we use the `/api/gnet` endpoint of the Grafana instance. To authenticate with that endpoint, the backend needs the following:
1. GCom API key (which we get from GCom via provisioning data)
2. Grafana instance Admin API key
3. URL of the Grafana instance

The Grafana API key is the source of most of the pain here. While it must be an Admin user that changes LLM settings, the backend plugin needs an Admin API key to act on the user's click, which it doesn't have. Just like with ML, the workaround is to use the frontend to create a (temporary, expires in 2 mins) API key and hand it to the backend.

(Note: we /could/ do this from the frontend, but passing the GCom key to the frontend is unsafe)

This is a rough attempt at the above.

Questions: do expired API keys vanish from the UI? Maybe the backend can delete the key when it is finished?


docker-compose.yaml
```diff
@@ -49,6 +50,31 @@ services:
...
+      - 'air'
+    command:
+      - '-c'
+      - '.air.llm-gateway.toml'
+    environment:
+      air_wd: /go/src/mlapi
+      OPENAI_API_KEY: '$OPENAI_API_KEY'
+      OPENAI_ORGANIZATION_ID: '$OPENAI_ORGANIZATION_ID'
+    restart: unless-stopped
+
```
csmarchbanks, line 77: 1. This repo is open source and others will not be able to do this.
2. The plugin will not be directly interacting with the llm-gateway, first it will go through the Grafana Cloud auth endpoint.

For local dev perhaps just directly use a Grafana Cloud deployment of llm-gateway specified via environment variables?

src/components/AppConfig/AppConfig.tsx
```diff
@@ -20,9 +20,35 @@ export interface LLMGatewaySettings {
   url?: string;
 }
 //////////////////////
+
```
manderson-dev, line 23: nit: you've added an api call function in the middle of a section that is declaring types/interfaces. There are a few other api calling functions at the bottom of the file, maybe move this down to that section of the file for a little more consistency.

src/components/AppConfig/AppConfig.tsx
```diff
@@ -20,9 +20,35 @@ export interface LLMGatewaySettings {
...
+  return lastValueFrom(
+    getBackendSrv().fetch({
+      url: `api/plugins/grafana-llm-app/resources/save-llm-state`,
+      method: 'POST',
+      data: { optIn },
+    })
+  )
+    .then((response: FetchResponse) => {
+      if (!response.ok) {
+        console.error(`Error using Grafana-managed LLM: ${response.status} ${response.data.message}`);
```
manderson-dev, line 34: One thing to consider, would be to show these failed results to the user via an alert. Not a blocker though.

manderson-dev (commented):
Thanks for tagging me on this. I ran this locally and did some testing on the settings page. Looked good

If we keep expanding this, it might be a good idea to create an /api folder and centralize api calls there, but for now I think the current setup is fine.

I was getting some errors around the backend service for the health check responding with a 503 after saving, and the health check loading spinner just kept spinning without indicating the health check had failed. I don't think that's in scope here but thought I'd mention it.

src/components/AppConfig/AppConfig.tsx
yoziru: Afaik these should pop up as an toast notification depending on the response: warning for 4xx, error for 5xx responses (at least on my end)

Screenshot 2024-02-07 at 15 58 47

pkg/plugin/resources.go
```diff
@@ -254,6 +275,133 @@ func (app *App) handleVectorSearch(w http.ResponseWriter, req *http.Request) {
...
+// [/api/gnet]: https://github.com/grafana/grafana/blob/4d8287b319514b750617c20c130ffc424a3ecf2c/pkg/api/api.go#L677
+// [proxies requests]: https://github.com/grafana/grafana/blob/4bc582570ef7e713599ab3f2009fa75c27bb8a02/pkg/api/grafana_com_proxy.go#L28
+// [`GrafanaComUrl` setting]: https://github.com/grafana/grafana/blob/460be702619428e455ba74f8fb3bb563c1bea43a/pkg/setting/setting.go#L1088
+// Handles requests to /save-llm-state endpoint, and pushes state to GCom.
+func (app *App) handleSaveLLMState(w http.ResponseWriter, req *http.Request) {
+	log.DefaultLogger.Debug("Handling request to save LLM state to gcom..")
+	if app.saToken == "" {
+		// not available in Grafana < 10.2.3 or if externalServiceAccounts feature flag is not enabled
+		log.DefaultLogger.Warn("Service account token not available; cannot save LLM state to gcom")
+		return
```
csmarchbanks, line 295: This should handle an error as well I think?

pkg/plugin/resources.go
```diff
@@ -254,6 +275,133 @@ func (app *App) handleVectorSearch(w http.ResponseWriter, req *http.Request) {
...
+		}
+	} else {
+		log.DefaultLogger.Warn("Request body is nil")
+	}
+
+	notHG := os.Getenv("NOT_HG")
+	if notHG != "" {
+		log.DefaultLogger.Info("NOT_HG variable found; skipping saving settings to gcom")
+		return
+	}
```
csmarchbanks, line 331: Could we do this check at the top of the function similar to the saToken check? I also think it needs error handling.

src/plugin.json
```diff
@@ -32,5 +32,12 @@
   "dependencies": {
     "grafanaDependency": ">=9.5.2",
```
csmarchbanks, line 33: Does the grafanaDependency need to be updated to 10.3? Or do we want to keep supporting earlier versions just without the service account and any features that depend on it?

pkg/plugin/resources.go
yoziru: In this case we want to return without an error, this is just the case where you’re running Grafana locally so we can early exit

src/plugin.json
yoziru: Hmm the service account is added in 10.2.3, but technically we’re backwards compatible in that the managed key code early exits and isn’t applicable anyway in those cases

manderson-dev (approved):
Will want someone else to thumbs up the go changes, but the ui work is good to go.

pkg/plugin/resources.go
csmarchbanks: I'm ok with that, but I guess my question is how would we get here if running Grafana locally? Seems like something would be wrong and a user should be warned?

pkg/plugin/resources.go
edwardcqian: Do we even need this logic? Isn't `NOT_HG` something we use for the ML plugin specifically

edwardcqian: Nvm, this logic is still needed for local dev. We need the option to not send to the gcom requests and just pass 200 so that the frontend save settings can pass without needing to set a valid gcom token

pkg/plugin/resources.go
edwardcqian: Updated this to just handle the error as well.

src/plugin.json
csmarchbanks: Tested out on an earlier version and the rest of the plugin works so let's keep it as is.

csmarchbanks (approved):

PR #216, opened 2024-02-09
Fixed llm-gateway communication details for Grafana Cloud

edwardcqian:
Fix a couple communication details with Grafana Cloud llm-gateway:
- health check unconfigured logic considers the `provider == grafana` case
- Added missing cortex gateway path prefix which was previous discarded (e.g. `https://<gateway-url>/llm`)
- Updated the API key and auth method

### DIFF
--- a/pkg/plugin/openai.go
+++ b/pkg/plugin/openai.go
@@ -11,7 +11,7 @@ import (
 	"path"
 )
 
-func (a *App) newAuthenticatedOpenAIRequest(ctx context.Context, method string, url url.URL, body io.Reader, tenant string) (*http.Request, error) {
+func (a *App) newAuthenticatedOpenAIRequest(ctx context.Context, method string, url url.URL, body io.Reader) (*http.Request, error) {
 	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url.String(), body)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


pkg/plugin/openai.go
```diff
@@ -22,7 +23,7 @@ func (a *App) newAuthenticatedOpenAIRequest(ctx context.Context, method string,
 	case openAIProviderAzure:
 		req.Header.Set("api-key", a.settings.OpenAI.apiKey)
 	case openAIProviderGrafana:
-		req.Header.Add("Authorization", "Bearer "+a.settings.LLMGateway.apiKey)
+		req.Header.Add("Authorization", "Basic "+a.settings.LLMGateway.apiKey)
```
csmarchbanks, line 26: Ugh, this does seem to be the case, but is also really confusing as we don't think of api keys as already having the tenant id combined with it. Could we either rename the apiKey field or separate them out to just be an api key and then use
`fmt.Sprintf("Bearer %s:%s", a.settings.Tenant, a.settings.LLMGateway.apiKey)`
or
`req.SetBasicAuth(a.settings.Tenant, a.settings.LLMGateway.apiKey)`?

edwardcqian: Updated to use `SetBasicAuth` where applicable

csmarchbanks (approved):

PR #219, opened 2024-02-15
Read LLM Allowed state from LLM-Gateway instead of gcom

csmarchbanks:


pkg/plugin/resources.go
```diff
@@ -452,13 +448,25 @@ func (app *App) handleSaveLLMOptInState(w http.ResponseWriter, req *http.Request
 
 	// write a success response body since backendSrv.* needs a valid json response body
 	w.WriteHeader(http.StatusOK)
-	_, err = w.Write([]byte(`{"status": "Success"}`))
+	_, _ = w.Write([]byte(`{"status": "Success"}`))
 	if err != nil {
 		handleError(w, fmt.Errorf("failed to write response body %w", err), http.StatusInternalServerError)
```
gerboland, line 453: Did you want to remove this branch then too?

src/components/AppConfig/AppConfig.tsx
```diff
@@ -61,34 +53,43 @@ export const AppConfig = ({ plugin }: AppConfigProps) => {
...
-    if (settings?.openAI?.provider === 'grafana' && !settings?.llmGateway?.isOptIn) {
+    if (settings?.openAI?.provider === 'grafana' && !managedLLMOptIn) {
       return "You must click the 'Enable OpenAI access via Grafana' button to use OpenAI provided by Grafana";
     }
     return;
   };
   const errorState = validateInputs();
 
+  const updateManagedLLMOptIn = (newOptIn: boolean): void => {
+    console.log('Try to set Opt in state to', managedLLMOptIn);
```
edwardcqian, line 72: Remove debug logs

src/components/AppConfig/AppConfig.tsx
```diff
@@ -61,34 +53,43 @@ export const AppConfig = ({ plugin }: AppConfigProps) => {
...
+    saveLLMOptInState(newOptIn);
```
edwardcqian, line 73: Do we need error handling here? In case saving the state fails.

pkg/plugin/resources.go
csmarchbanks: Ahh yep. After writing a success header you can't re-write the header to a failure so not much we can do.

src/components/AppConfig/AppConfig.tsx
csmarchbanks: Yeah, I just changed our fetch functions to throw the error which gives you a nice popup and will no longer update the state.

gerboland (commented):
Looks good, but since I authored a big chunk, I need someone else to approve.

edwardcqian (approved):

PR #220, opened 2024-02-16
Added logic to save plugin config to gcom provisioned plugin

edwardcqian:


pkg/plugin/resources.go
```diff
@@ -341,73 +400,175 @@ func (app *App) handleSaveLLMState(w http.ResponseWriter, req *http.Request) {
...
+	// Read the request body
+	if req.Body == nil {
+		log.DefaultLogger.Warn("Request body is nil")
+		handleError(w, errors.New("request body required"), http.StatusBadRequest)
+		return
+	}
+
+	log.DefaultLogger.Debug("Getting provisioned plugin ID from grafana.com")
+
+	notHG := os.Getenv("NOT_HG")
```
csmarchbanks, line 516: We shouldn't require OSS users to set `NOT_HG` to make this work, and should instead either use our existing flag in the settings, or look for something like `HG_INSTANCE_SLUG` as the environment variable. If choosing the latter check with the Hosted Grafana team to identify the best environment variable to use.

src/components/AppConfig/AppConfig.tsx
```diff
@@ -202,6 +203,25 @@ export const updatePlugin = (pluginId: string, data: Partial<PluginMeta>) => {
...
+
+  return lastValueFrom(response);
+};
+
+export const updateAndSavePluginSettings = (pluginId: string, data: Partial<PluginMeta>) => {
+  const gcomPluginData = {
+    jsonData: data.jsonData,
+    secureJsonData: data.secureJsonData,
+  };
+  return Promise.all([updatePlugin(pluginId, data), savePluginSettings(gcomPluginData)]);
```
csmarchbanks, line 222: 🤔 Pros/cons of doing two calls from the client vs using the backend to call both places? Probably doesn't matter and this is fine to start with?

sd2k: Unless I'm misunderstanding, the frontend won't have access to `secureJsonData`, it's only available in the backend (otherwise it wouldn't be very secure!)

pkg/plugin/resources.go
edwardcqian: I added the NOT_HG to attempt to make local dev work.

Checked a dev hg instance, it should be `HG_INSTANCE_SLUG`, will use this instead.

src/components/AppConfig/AppConfig.tsx
edwardcqian: The secureJsonData from the frontend would be everything that is updated upon save.

pkg/plugin/resources.go
```diff
@@ -467,6 +469,107 @@ func (a *App) handleLLMState(w http.ResponseWriter, req *http.Request) {
...
+	var err error
+	if hgSlug != "" && a.settings.EnableGrafanaManagedLLM {
+		pluginID, err = getPluginID(req.Context(), a.settings.Tenant, a.grafanaAppURL, a.saToken, a.settings.GrafanaComAPIKey)
+		if err != nil {
+			handleError(w, fmt.Errorf("get plugin ID: %w", err), http.StatusInternalServerError)
+			return
+		}
+	} else {
+		log.DefaultLogger.Info("Hosted Grafana Slug not found or plugin not provisioned; skipping getting provisioned pluginID")
+		pluginID = 0
```
csmarchbanks, line 528: Could we write success and return here rather than requiring a second check of the setting and environment variable below?

If so typical go practice would be to invert the logic to do the if check and bail to avoid as much nesting of if statements.
```
if hgSlug == "" || !a.settings.EnableGrafanaManagedLLM {
    w.WriteHeader(http.StatusOK)
    _, _ = w.Write([]byte(`{"status": "success"}`))
    return
}
...

File: pkg/plugin/resources.go [hunk 4]
```diff
@@ -467,6 +469,107 @@ func (a *App) handleLLMState(w http.ResponseWriter, req *http.Request) {
...
+	if err != nil {
+		handleError(w, fmt.Errorf("create gcom request: %w", err), http.StatusInternalServerError)
+		return
+	}
+	gcomReq.Header.Set("Authorization", fmt.Sprintf("Bearer %s", a.saToken))
+	gcomReq.Header.Set("X-Api-Key", a.settings.GrafanaComAPIKey)
+	gcomReq.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+
+	if hgSlug != "" && a.settings.EnableGrafanaManagedLLM {
+		_, err = doRequest(gcomReq)
```
csmarchbanks, line 542: Error handling is necessary here

pkg/plugin/resources.go
```diff
@@ -467,6 +469,107 @@ func (a *App) handleLLMState(w http.ResponseWriter, req *http.Request) {
...
+		log.DefaultLogger.Info("Hosted Grafana Slug not found or plugin not provisioned; skipping saving settings to grafana.com", "Request details:", fmt.Sprintf(" %+v", gcomReq))
+	}
+
+	// write a success response body since backendSrv.* needs a valid json response body
+	w.WriteHeader(http.StatusOK)
+	_, _ = w.Write([]byte(`{"status": "Success"}`))
+	if err != nil {
+		handleError(w, fmt.Errorf("failed to write response body %w", err), http.StatusInternalServerError)
+		return
+	}
```
csmarchbanks, line 553: This block can't really do much after we already write a 200 header so can be removed.

pkg/plugin/resources.go
```diff
@@ -467,6 +469,105 @@ func (a *App) handleLLMState(w http.ResponseWriter, req *http.Request) {
...
 
+// provisionedPlugin is the response returned by a call to grafana.com's provisioned-plugin's endpoint
+// for an specific instance's plugin.
+type provisionedPlugin struct {
+	ID int `json:"id"`
+}
+
+// getPluginID gets the *ID* of the *provisioned plugin* from grafana.com.
+// Note that this differs to the plugin ID referred to by the `backend.CallResourceRequest`,
+// which is 'grafana-ml-app'
```
csmarchbanks, line 480:
```suggestion
// which is 'grafana-llm-app'
```

pkg/plugin/resources.go
```diff
@@ -467,6 +469,105 @@ func (a *App) handleLLMState(w http.ResponseWriter, req *http.Request) {
...
+	}
+
+	pluginID, err = getPluginID(req.Context(), a.settings.Tenant, a.grafanaAppURL, a.saToken, a.settings.GrafanaComAPIKey)
+	if err != nil {
+		handleError(w, fmt.Errorf("get plugin ID: %w", err), http.StatusInternalServerError)
+		return
+	}
+
+	gcomPath := fmt.Sprintf("/api/gnet/instances/%s/provisioned-plugins/%d", a.settings.Tenant, pluginID)
+	gcomReq, err := http.NewRequestWithContext(req.Context(), "POST", a.grafanaAppURL+gcomPath, req.Body)
```
csmarchbanks, line 534: Will we need to merge any existing `secureJsonData` before making this request? I wouldn't want something like our provisioned api key to be overwritten.

pkg/plugin/resources.go
```diff
@@ -467,6 +469,106 @@ func (a *App) handleLLMState(w http.ResponseWriter, req *http.Request) {
...
+	// The HG_INSTANCE_SLUG is only set for Grafana Cloud instances.
+	hgSlug := os.Getenv("HG_INSTANCE_SLUG")
+	var pluginID int
+	var err error
+
+	if hgSlug == "" || !a.settings.EnableGrafanaManagedLLM {
+		log.DefaultLogger.Info("Hosted Grafana Slug not found or plugin not provisioned; skipping saving settings to grafana.com")
+		w.WriteHeader(http.StatusOK)
+		_, _ = w.Write([]byte(`{"status": "Success"}`))
+	}
```
csmarchbanks, line 525: Need a return after the write to avoid continuing to gcom related code.

src/components/AppConfig/AppConfig.tsx
gerboland: I think Ben's point is that the frontend shouldn't be even trying to write to secureJsonData. 

I noticed there's a `Secrets` type on line 25 already, that is supposedly populated from the initial settings. What makes them secure is that the backend does not pass the secrets to the frontend at all, only a "true" value - see here for a working-ish dev instance:
image

I think you can avoid trying to push secureJsonData from the frontend.

csmarchbanks: This has now been updated a bit but the flow is:
1. Update gcom by calling an endpoint in the plugin with all jsonData and any new secureJsonData. The endpoint needs to merge the new secureJsonData with the old in order to not lose data in gcom.
2. Update the plugin settings in Grafana (which will then be pulled from gcom on the next restart). This is the normal plugin update and doesn't need any merging as Grafana handles that for us.

I don't love that the state has to be sent to both of those places, but I don't see a way around it either. Anyone else have a better idea?

pkg/plugin/resources.go
```diff
@@ -464,6 +466,154 @@ func (a *App) handleLLMState(w http.ResponseWriter, req *http.Request) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshal request body to JSON %w", err)
+	}
+
+	// Update mandatory fields
+	requestData.SecureJSONData[encodedTenantAndTokenKey] = base64.StdEncoding.EncodeToString([]byte(a.settings.DecryptedSecureJSONData[encodedTenantAndTokenKey]))
+
+	// Insert existing plugin secureJSONData fields if missing from request
+	for key, value := range a.settings.DecryptedSecureJSONData {
+		if existingValue, exists := requestData.SecureJSONData[key]; !exists || existingValue == "" {
```
csmarchbanks, line 524: Should this only check `exists`? Otherwise there would be no way to unset a piece of secure json data by setting it to ""?

csmarchbanks (commented):
👍 I am happy to get this deployed to dev to test further and tweak as needed. I would like someone else to approve as well though.

sd2k (commented):
This looks good to me except a couple of console logs left in there!

gerboland (approved):
No notes, aside from what Ben found. Let's get this in dev!

PR #222, opened 2024-02-16
Settings UI v2

gerboland:
Quick second iteration of the Settings UI, taking some design feedback into account and simplifying the flow:
- no more dialogs showing long terms & conditions, or opt-out message

Grafana-managed option:
image

Own OpenAI a/c option:
image


csmarchbanks (approved):
Seems to be working well for me! I agree that the no-modal flow does feel a bit more natural and like that there are not multiple blue buttons to click.

PR #228, opened 2024-02-19
Update licence terms

gerboland:
- Update our license terms from legal.
- Add workaround for being unable to click links inside the Card (and fixes scrolling the license terms too)

### DIFF
--- a/src/components/AppConfig/LLMConfig.tsx
+++ b/src/components/AppConfig/LLMConfig.tsx
@@ -89,66 +89,99 @@ export function LLMConfig({
 
   return (
     <FieldSet label="OpenAI Settings" className={s.sidePadding}>
+      {' '}
       {allowGrafanaManagedLLM && (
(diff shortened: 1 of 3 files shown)


src/components/AppConfig/LLMConfig.tsx
```diff
@@ -89,59 +89,105 @@ export function LLMConfig({
 
   return (
     <FieldSet label="OpenAI Settings" className={s.sidePadding}>
+      {' '}
```
csmarchbanks, line 92: What's this for?

csmarchbanks (approved):
One small question but 👍

src/components/AppConfig/LLMConfig.tsx
gerboland: bah, nothing. My editor really loves preserving whitespace like that

PR #235, opened 2024-02-22
Dependency improvement

SandersAaronD:
Intended to accomplish the following (all of which I tried to do individually and could not feasibly do without doing the other things):
1) Move dependencies from @grafana/experimental into a monorepo with the plugin as a new npm package @grafana/llm
2) Set up build and deployment for the restructured monorepo for both the plugin and those dependencies
3) Set up a development sandbox that allows testing changes to @grafana/llm directly as they are made

Todo to make this viable:

- [x] Make sure installing from root, building from root, and testing from root all work
- [x] Set up CI/CD
- [x] Update documentation to reflect new method of running the module

Current iteration as of right now is perfectly runnable, but you have to navigate to subdirectories to run things well. This is not ideal.


SandersAaronD:
1) Jest fails to properly resolve workspace dependencies unless you add jest to root instead of running it from each project,
2) Migrating jest to root means that now when some tests run we get an error that I think is multiple copies of react ending up in scope for jest,
3) Migrating react to the root scope just breaks everything in new and different ways (not currently reflected on branch, it seemed extremely debatable if this was an improvement)
4) Drone is failing this pr for some reason that I am pretty sure is a problem for much later.

SandersAaronD:
Owing to the size of the refactor here and the fact that other PRs keep slowing down progress, I think it is probably smarter to make sure that the current state of the repo is documented (so other people can build and release) and to put CI/CD into a separate ticket. We can do manual build + release until CI/CD is set up, which should not be extremely long, and in the meantime I will not have to rebase constantly (and hope the changes don't end up generating more work).

SandersAaronD:
Everything seems good to me, so I'm marking this ready for review.

This still needs a deployment key added to its secrets, and then we need to test the release flow's final step (pushing the plugin out). We could test this some off of this branch, but since it's a branch-sensitive operation (ie, meant to be triggered in a certain way off of specifically main) I don't think we can be sure it works correctly until we actually run it off of main.

I'd be comfortable merging this in, subject to getting the secret added.

SandersAaronD:
Secret is added, separate script is set up to do OS/arch-specific builds. Does need testing for a "live deploy" but looks 100% clean to me, this should port the repository to github actions cleanly in addition to the other work.

README.md
```diff
@@ -336,27 +102,12 @@ The `.subscribe` method can take [a few different forms](https://github.com/Reac
    npm run lint:fix
    ```
 
-### Developing with the Example App
```
edwardcqian, line 339: Might be nice to add a reference to `DevSandbox` here.

README.md
```diff
@@ -336,27 +102,12 @@ The `.subscribe` method can take [a few different forms](https://github.com/Reac
...
-- Add the following volume:
-```
-<some-parent-path>/grafana-llm-app/dist:/var/lib/grafana/plugins/grafana-llm-app
-```
-3. Follow the instructions in the llm example app to run the app
-
 ## Release process
 
 ### Plugin Release
 - Bump version in package.json (e.g., 0.2.0 to 0.2.1)
```
edwardcqian, line 108:
```suggestion
- Bump version in package.json (e.g., 0.2.0 to 0.2.1)
- Bump version in `packages/grafana-llm-app/package.json` (e.g., 0.2.0 to 0.2.1)
```

edwardcqian (approved):
Able to build and run the plugin and llm package.

Tested the new llm package via dev sandbox

README.md
```diff
@@ -336,27 +102,12 @@ The `.subscribe` method can take [a few different forms](https://github.com/Reac
...
-1. Clone the [llm example app](https://github.com/grafana/grafana-llmexamples-app/tree/main)
-2. Update the following fields in `docker-compose.yaml` in the llm example app
-- comment out # GF_INSTALL_PLUGINS: grafana-llm-app
-- Add the following volume:
-```
-<some-parent-path>/grafana-llm-app/dist:/var/lib/grafana/plugins/grafana-llm-app
-```
-3. Follow the instructions in the llm example app to run the app
-
 ## Release process
```
edwardcqian, line 105: Do we need a release process the `@grafana/llm`?

SandersAaronD: We do; I want to probably add that separately, I'll open an issue for it.

README.md
SandersAaronD: Good idea, I'll add that.

SandersAaronD:
> Able to build and run the plugin and llm package.
> 
> Tested the new llm package via dev sandbox.
> 
> How does the new release process work with github actions?

We should be able to trigger the release from actions after a merge to main with a button click; this part is tricky to test until the PR is merged in.

Issue #241, opened 2024-02-27
Config UI: changing LLM option & save, then switch tab away & back, old option still selected


gerboland:
In the Config page:
1. Take note of the LLM option selected
2. Change the LLM selection to another option and click Save & Test
3. Click the Overview Tab
4. Click the Configuration Tab

Bug: the LLM option at step 1 is selected. But if I reload the page, the correct option is then shown.

The issue if the `jsonData` provided to the plugin is not updated

sd2k:
Was fixed by #242 I believe

PR #242, opened 2024-02-27
Healthcheck fix when disabled, UI bug fixes

gerboland:
This is a bunch of bug fixes:
1. if LLMs was disabled in the plugin config, the health check could still return a enabled LLM if a previous OpenAI API key was provided. The default state in the loadSettings() was the culprit.
2. Fix for bug #241. The plugin relies on the `jsonData` provided to it by Grafana to know its current state (read from the backend on plugin initialisation). If the config page changes the configuration, the plugin `jsonData` remains unchanged until a page reload happens. Fixed by pushing the new jsonData to the plugin on save.
3. Removal of some state duplication in LLMConfig (setLLMOption state was unnecessary)
4. Clear the health check results if user changes any setting.
5. I tightened up the health check (empty api key only permitted if llms disabled or grafana-managed).

### DIFF
--- a/pkg/plugin/health.go
+++ b/pkg/plugin/health.go
@@ -88,7 +88,7 @@ func (a *App) openAIHealth(ctx context.Context, req *backend.CheckHealthRequest)
 
 	d := openAIHealthDetails{
 		OK:         true,
-		Configured: a.settings.OpenAI.apiKey != "" || a.settings.OpenAI.Provider == openAIProviderGrafana,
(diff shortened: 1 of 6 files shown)


gerboland:
> Not sure how to test 4.

Select the own-OpenUI option, hit Save & Test - the health check result should appear. Then click the managed-LLM option. The health check result should vanish.

> 5 might cause issues with self-hosted openai compatible service that doesn't require an api key, but maybe we want a separate way to handle those anyways.

Yep, that's a whole different problem.

edwardcqian (approved):

PR #243, opened 2024-02-27
Retry requests to gcom that 409

gerboland:
GCom occasionally returns a 409 when under load, this adds a simple retry which typically helps.


csmarchbanks (approved):
Annoying that it is necessary but 👍

PR #249, opened 2024-03-07
fix: make sure the Grafana OpenAI proxy works when hosted at a subpath

sd2k:
Prior to this commit we ignored the path of the configured OpenAI API
when modifying the proxied URL of a request, so that even if the OpenAI
API URL was set to something like "https://foo.com/llm" we would still
proxy requests to "https://foo.com".

This commit preserves the path of the configured URL so that subpaths
are honoured in the outgoing proxied request from the plugin.

I also fixed a missing early return after failing to modify the URL.

Fixes https://github.com/grafana/incident/issues/5107.

### DIFF
--- a/pkg/plugin/resources.go
+++ b/pkg/plugin/resources.go
@@ -48,7 +48,14 @@ func modifyURL(openAIUrl string, req *http.Request) error {
 	}
 	req.URL.Scheme = u.Scheme
 	req.URL.Host = u.Host
-	return nil
+	// At this point:
+	// - u.Path is something like '' or '/llm', depending on the OpenAI provider.
+	// - req.URL.Path is something like '/openai/v1/chat/completions', since we require the
+	//   '/openai' prefix to be present in the request URL.
+	// We want the final URL to be something like '/llm/openai/v1/chat/completions'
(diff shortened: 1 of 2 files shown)


pkg/plugin/resources.go
```diff
@@ -202,7 +210,7 @@ type grafanaOpenAIProxy struct {
 }
 
 func (a *grafanaOpenAIProxy) ServeHTTP(w http.ResponseWriter, req *http.Request) {
-	err := modifyURL(a.settings.LLMGateway.URL+"/openai", req) // GER: FIXME - not durable to / added
```
jtheory, line 205: Fixes all look valid to me; I'm just confused by why `+"/openai"` was added in here -- it's not actually used in modifyURL(), right? So I'm guessing this is just an artifact of earlier experimental code

sd2k: Yeah it's been ignored by `modifyURL` for a while I think, looks like an artifact.

jtheory (approved):
Simple logic change, does what it says on the tin, and adds test coverage for the now-supported flow. Also adds a missing return after an error that was missing.

PR #259, opened 2024-03-11
Fix validate step again

SandersAaronD:
Permissions were bad on vault and the vault thing needed a tweak


gerboland (approved):
looks sane

PR #272, opened 2024-03-14
Update to react-router-v6

edwardcqian:


sd2k:
I'm not even sure we need to bother using react-router here, we only have one page 🤷 could just return  `<Models />`?

edwardcqian:
> I'm not even sure we need to bother using react-router here, we only have one page 🤷 could just return `<Models />`?

Removed the whole App bit since we don't really need it, but might be a bit strange to set the root page directly to Models. Thoughts? 🤔

sd2k (approved):

PR #274, opened 2024-03-14
Change p to div inside cards

edwardcqian:
Not sure if this is the right read, seems like something changed within the cards components that is complaining that we are using nested `<p>`

Changing to <div> only slightly changes the formatting, hopefully this is okay 🤔

### DIFF
--- a/packages/grafana-llm-app/src/components/AppConfig/LLMConfig.tsx
+++ b/packages/grafana-llm-app/src/components/AppConfig/LLMConfig.tsx
@@ -105,17 +105,17 @@ export function LLMConfig({
           >
             <Card.Heading>Use OpenAI provided by Grafana</Card.Heading>
             <Card.Description>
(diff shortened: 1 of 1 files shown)


csmarchbanks (approved):

SandersAaronD:
I think this is actually a dependency issue fwiw; I saw it during the (long) saga of trying to get stuff refactored correctly, and am not sure what combination of upgrades/tweaks made this come up again. If this change fixes it it's worth merging in, although worth noting that it doesn't _seem_ to break anything, it just throws warnings etc.

edwardcqian:
> I think this is actually a dependency issue fwiw; I saw it during the (long) saga of trying to get stuff refactored correctly, and am not sure what combination of upgrades/tweaks made this come up again. If this change fixes it it's worth merging in, although worth noting that it doesn't _seem_ to break anything, it just throws warnings etc.

CI's failing because of it though.

PR #282, opened 2024-03-18
Only apply saveLLMOptInState when enableGrafanaManagedLLM is enabled

edwardcqian:
Currently `saveLLMOptInState` is called even when `enableGrafanaManagedLLM` is not set.

This causes issues with OSS version of this plugin since the llm-gateway call will fail.


sd2k (approved):
LGTM

PR #288, opened 2024-03-25
add LLM app plugin keywords

jtheory:
The new plugins search uses keywords (if you set them):
https://grafana.com/docs/grafana-cloud/whats-new/#plugin-search-improvements

Any others I missed?


sd2k (approved):
Good find!

PR #289, opened 2024-03-25
Add gcom call to release action, bump version so we can run a release

SandersAaronD:
Does what it says on the tin, isn't ready to run a release from until we have GCOM_ROOT in secrets, which it isn't presently. I am, I think, as sure as I can be that this will work smoothly once that's populated, and there's no real way to be sure until we do it. (Also, until we do we're going to have to do public releases of this plugin by calling gcom manually).

Hopefully we see these actions get merged into shared-workflows somewhere down the line and we can swap this to not reference my personal repository.

New action referenced is here: https://github.com/grafana/shared-workflows/pull/73


.github/workflows/plugin-release.yml
```diff
@@ -194,4 +199,15 @@ jobs:
         with:
           path: ./${{ needs.build-release.outputs.upload-folder }}
           destination: 'integration-artifacts/grafana-llm-app/release/'
-          parent: false
\ No newline at end of file
+          parent: false
+
+      - id: 'publish-to-gcom'
+        name: 'Publish to Grafana.com'
+        uses: 'sandersaarond/shared-workflows/actions/publish-backend-plugin-on-site@publish-plugin-on-site'
```
gerboland, line 206: I'm missing insights as I cannot find this code. What is this doing exactly?

It's not doing a release ([updating a file here](https://github.com/grafana/deployment_tools/tree/d04a0910618784758a78d80d7041838eec3fc86c/ksonnet/environments/hosted-grafana/waves/provisioned-plugins/grafana-llm-app)), is it?

gerboland: I'm mainly concerned with the release process. I'm probably echoing your thinking by saying I would prefer we do an internal release to cloud (dev/staging/ops) first, validate it is happy, before pushing it to both cloud prod & the gcom plugins list.

I'd love a dev non-hosted grafana instance with the plugin installed that we could check is functional, to test OSS user's experience.

SandersAaronD: Yes, sorry -- this is making sure to add the newer version of the plugin to the Grafana plugins list for OSS. I agree that I would want that, but I think it's a todo for the future. Right now this is just making it to parity with the pre-existing Drone workflow, which I do not think had any kind of dev/staging set up (unfortunately).

gerboland: Ok, fair. If something goes wrong, is it clear how do we perform a revert?

SandersAaronD: Lost this thread apparently, but:

We'd have to do another minor or patch version bump and re-release for a bad plugin version. gcom releases are not, i think, really reversible.

In the event that this PR specifically is bad I think it's easy to revert, the changes are relatively minor.

PR #315, opened 2024-05-03
Abstract models to allow more providers

csmarchbanks:
Rather than having the client pass in OpenAI specific models create an abstraction for a default and a high accuracy model. For backwards compatability gpt-3.5-turbo is mapped to default and gpt-4 is mapped to high accuracy.

As part of this it is also required to read the body in order to transform from an abstracted model to the specific model (or deployment in the case of Azure) being used. A simple interface is used that currently only supports Models and ChatCompletions, but can be expanded to support StreamChatCompletions in the future (and refactor the streaming code). This also means we now limit the endpoints that our API allows to just be Models and ChatCompletions as well. These are all that we have really seen used so it should be ok.

I have also avoided updating the frontend to use these new model names to make sure that this is truly backwards compatible. We can update the frontend in a future PR after this has been tested and merged. Ideally deployed as well as a frontend updating before a backend could cause breakages.

### DIFF
--- a/packages/grafana-llm-app/llmclient/llmclient.go
+++ b/packages/grafana-llm-app/llmclient/llmclient.go
@@ -19,6 +19,23 @@ const (
 	appResourcesPrefix = appPrefix + "/resources"
 )
 
+// Model is an abstraction over the different models available in different providers
+type Model string
+
+const (
+	// ModelSmall is the small model, which is the fastest and cheapest to use. OpenAI default: gpt-3.5-turbo
+	ModelSmall = "small"
+	// ModelMedium is the medium model, which is a good balance between speed and cost. OpenAI default: gpt-4-turbo
+	ModelMedium = "medium"
+	// ModelLarge is the large model, which is the most powerful and accurate. OpenAI default: gpt-4
+	ModelLarge = "large"
+)
+
+type ChatCompletionRequest struct {
+	openai.ChatCompletionRequest
+	Model Model `json:"model"`
+}
+
 // OpenAI is an interface for talking to OpenAI via the Grafana LLM app.
 // Requests made using this interface will be routed to the OpenAI backend
 // configured in the Grafana LLM app's settings, with authentication handled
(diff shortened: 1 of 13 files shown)


packages/grafana-llm-app/pkg/plugin/llm_provider.go
```diff
@@ -0,0 +1,129 @@
...
+	"strings"
+)
+
+var errBadRequest = errors.New("bad request")
+
+type Model string
+
+const (
+	ModelDefault      = "default"
+	ModelHighAccuracy = "high-accuracy"
```
sd2k, line 16: Rather than having a model named 'default' would it be better to have slightly more descriptive names and allow the user to change the default as they see fit? We could still set the default to the 'mid range' model to begin with.

Maybe this is just unnecessary complexity though - would users actually want to do that?

(I realise now this should have been in the design doc, sorry!)

yoziru: 👍 yep, to keep things more in line with existing semantics, we're swapping this to small/medium/large

packages/grafana-llm-app/llmclient/llmclient.go
```diff
@@ -19,6 +19,23 @@ const (
...
+const (
+	// ModelSmall is the small model, which is the fastest and cheapest to use. OpenAI default: gpt-3.5-turbo
+	ModelSmall = "small"
+	// ModelMedium is the medium model, which is a good balance between speed and cost. OpenAI default: gpt-4-turbo
+	ModelMedium = "medium"
+	// ModelLarge is the large model, which is the most powerful and accurate. OpenAI default: gpt-4
+	ModelLarge = "large"
+)
+
+type ChatCompletionRequest struct {
```
sd2k, line 34:
```suggestion
// ChatCompletionRequest is a request for chat completions using an abstract model.
type ChatCompletionRequest struct {
```

packages/grafana-llm-app/pkg/plugin/grafana_provider.go
```diff
@@ -0,0 +1,64 @@
...
+
+func (p *grafanaProvider) Models(ctx context.Context) (ModelResponse, error) {
+	return ModelResponse{
+		Data: []ModelInfo{
+			{ID: ModelSmall},
+			{ID: ModelMedium},
+			{ID: ModelLarge},
+		},
+	}, nil
+}
```
sd2k, line 38: Does the gateway currently support these models?

yoziru: Hmm gateway should act as an OpenAI server e.g. not care about our abstractions. Our plugin backend should translate these abstractions to their respective actual model names before making the request to gateway.

Not sure what we want to do with this endpoint yet, afaik it's not in use apart from a test page 🤔 
I think we will want to fetch the actual models that are available on the API (this applies to any provider, gateway should just send it through). This way when we expose the abstraction in settings, we can provide a drop-down to the user, based on which models are available.

sd2k: Yeah makes sense, I didn't think we used it anywhere yet.

sd2k (approved):
Happy with this as a first go! Let's get the streaming stuff moved over soon too.

PR #317, opened 2024-05-03
Update SA token retrieval logic

edwardcqian:
Grafana external service account tokens should be retrieved via the plugin sdk.

### DIFF
--- a/packages/grafana-llm-app/pkg/plugin/app.go
+++ b/packages/grafana-llm-app/pkg/plugin/app.go
@@ -2,6 +2,7 @@ package plugin
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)


packages/grafana-llm-app/pkg/plugin/app.go
```diff
@@ -71,7 +72,12 @@ func NewApp(ctx context.Context, appSettings backend.AppInstanceSettings) (insta
 	app.CallResourceHandler = httpadapter.New(mux)
 
 	// Getting the service account token that has been shared with the plugin
-	app.saToken = os.Getenv("GF_PLUGIN_APP_CLIENT_SECRET")
+	cfg := backend.GrafanaConfigFromContext(ctx)
+	app.saToken, err = cfg.PluginAppClientSecret()
+	if err != nil {
+		log.DefaultLogger.Error("Error getting service account token")
+		return nil, fmt.Errorf("error getting service account token: %w", err)
```
yoziru, line 79: This is a breaking change if your Grafana instance doesn't have the `externalServiceAccounts` feature toggle turned on. Think we should log a warning and not return an error here. Or we could also first check if that feature is toggled on before trying to get the SA token this way

yoziru (changes requested):

packages/grafana-llm-app/pkg/plugin/app.go
edwardcqian: Hmm, that's a good point. Will change to warning.

yoziru (approved):

PR #325, opened 2024-05-07
minor dev setup improvements

yoziru:
- docker compose: only bring up Grafana by default
- enable server router logging
- add a `npm run backend:restart` command to rebuild and restart the backend plugin without having to bring the container up/down


sd2k (approved):
:pray:

PR #328, opened 2024-05-08
testing: add configurable test provider for use in integration tests

sd2k:
This adds a new LLMProvider which just returns some messages defined in the JSONData. It's configured by setting `settings.OpenAI.Provider` to "test". This will allow us to test the various other plugin aspects (resource calls, streaming calls) in a more realistic environment without having to configure OpenAI/Azure or run a separate LLM gateway.

### DIFF
--- a/packages/grafana-llm-app/provisioning/plugins/test-provider/grafana-llm-app.yaml
+++ b/packages/grafana-llm-app/provisioning/plugins/test-provider/grafana-llm-app.yaml
@@ -1,0 +1,74 @@
+apiVersion: 1
+
+apps:
+  - type: grafana-llm-app
+    jsonData:
+      base64EncodedAccessTokenSet: True
+      # enableGrafanaManagedLLM: True
+      # displayVectorStoreOptions: False
+      openAI:
+        provider: test
+        # url: https://api.openai.com
+        # organizationId: $OPENAI_ORGANIZATION_ID
+        testProvider:
+          modelsResponse: 
+            data:
+            - id: small
+            - id: medium
+            - id: large
+
+          chatCompletionResponse:
+            id: "1234"
+            model: "small"
+            choices:
+            - message:
+                role: "assistant"
+                content: "How are you today?"
+          chatCompletionError: ""
+
+          initialStreamError: ""
+          streamDeltas:
+            - role: "assistant"
+              content: "How "
+            - role: "assistant"
+              content: "are "
+            - role: "assistant"
+              content: "you "
+            - role: "assistant"
+              content: "today"
+            - role: "assistant"
+              content: "?"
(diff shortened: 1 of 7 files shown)


csmarchbanks (commented):
Generally 👍 on the idea, could you add a bit of verification for the request in each function though? Make sure a prompt and valid model is passed is my idea to make sure the request side is also working properly.

packages/grafana-llm-app/provisioning/plugins/test-provider/grafana-llm-app.yaml
```diff
@@ -0,0 +1,69 @@
...
+      # displayVectorStoreOptions: False
+      openAI:
+        provider: test
+        # url: https://api.openai.com
+        # organizationId: $OPENAI_ORGANIZATION_ID
+        testProvider:
+          modelsResponse: 
+            data:
+            - id: small
+            - id: medium
```
csmarchbanks, line 17:
```suggestion
            - id: base
```

packages/grafana-llm-app/provisioning/plugins/test-provider/grafana-llm-app.yaml
```diff
@@ -0,0 +1,69 @@
...
+            - id: large
+
+          chatCompletionResponse:
+            id: "1234"
+            model: "small"
```
csmarchbanks, line 22:
```suggestion
            model: "base"
```

csmarchbanks (commented):
Running into some issues trying this out, I think around the health check and frontend looking for specific configuration fields/not knowing about the test provider.

sd2k:
Yep it looks like some things were broken after the last few PRs. I've rebased and added some fixes so :crossed_fingers: it's working now!

packages/grafana-llm-app/src/components/AppConfig/LLMConfig.tsx
```diff
@@ -221,6 +235,14 @@ export function LLMConfig({
             </Card.Figure>
           </Card>
         </div>
+        {(llmOption === 'test' || previousOpenAIProvider === 'test') && (
```
csmarchbanks, line 238: Maybe display this when `process.env.NODE_ENV === 'development'` so that it isn't required to know about provisioning it to use test mode in dev?

csmarchbanks (approved):
One small idea, but 👍 anyway.

packages/grafana-llm-app/src/components/AppConfig/LLMConfig.tsx
sd2k: Nice idea :+1: It still feels weird to me using environment variables in frontend code, I forget that it works!

PR #330, opened 2024-05-09
Use go-openai for ChatCompletion requests

csmarchbanks:
Now that we depend on go-openai we can use it for ChatCompletion (non streaming) requests as well to remove our manual HTTP calls. This also reduces the number of structs we define, preferring those in go-openai when we are not overriding a type, such as with Model.


sd2k (approved):
Tested all three providers and LGTM :+1:
Love to see it:
image

:shipit:

PR #331, opened 2024-05-09
Abstracted model frontend config

sd2k:
Creating this on behalf of @yoziru since he's offline today.

This PR:
- ~updates abstract model names from small/medium/large to base/large~ split into #334 
- adds a backend setting to configure the mapping between these abstract names to concrete names understood by the provider (base -> gpt-3.5-turbo, etc)
- updates the frontend settings page, adding a section to the form where users can configure:
  - the default model, used in case no model is specified or in case the specified model is not found
  - overrides for the mapping from abstract to concrete names

Still TODO before review/merge:

- [x] double check this works as expected for Azure
  - [x] do we want users to specify mappings from `<abstract> -> <openai model>`, or `<abstract> -> <azure deployment>`?
    - I ended up keeping the `<abstract> -> <azure deployment>` which we've always had in the Azure settings, and hiding the `<abstract> -> <openai model>` field if Azure is selected, since we don't use it anyway. This is one fewer thing for the user to have to think about.
  - [x] I don't think it's working when the model is not specified yet, need to check that
- [x] decide what to do about an empty 'model' field for Grafana provider
  - [x] currently the backend panics which is no good, need to fix that


sd2k:
Rebased onto #335 and health checks look good now.

yoziru:
Can close this now, all of this has been broken up and merged in separate PRs

PR #334, opened 2024-05-10
model abstraction: S/M/L -> Base/Large

yoziru:
Better naming + simpler is better
Diff between medium and large isn't clear for devs, and small would lead devs to maybe not use (and steer towards middle). Instead use Base/Large, so it's clear most use cases should try to use Base (or not define a model abstraction at all), and only use Large when you need that large context / more advanced capabilities.

This is breaking if any use cases already switched to small/medium before, but not handling those because
- 0.9.x was just released and was mostly a backend thing
- we'll handle fallback to default when we expose the frontend settings (so this will go into 0.10.x)

Split out from https://github.com/grafana/grafana-llm-app/pull/331 to make that PR smaller

### DIFF
--- a/packages/grafana-llm-app/README.md
+++ b/packages/grafana-llm-app/README.md
@@ -73,7 +73,8 @@ apps:
         provider: azure
         url: https://<resource>.openai.azure.com
         azureModelMapping:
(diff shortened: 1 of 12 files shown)


sd2k (approved):
LGTM :+1: nice to split it up

PR #335, opened 2024-05-10
Use go-openai for health check requests

sd2k:
This completes the migration away from using custom clients
and uses the configured provider for health checks too.


yoziru (approved):
LGTM ✅

PR #338, opened 2024-05-10
Add model config to backend, with default and mappings between abstract and concrete models

sd2k:
This is split out of #331 and only contains the squashed backend changes.


yoziru (approved):
LGTM 🚀

Issue #323, opened 2024-05-07
Expose abstract models in front-end config to allow overrides and custom models


yoziru:
- Expose the model abstractions in the front-end
- Have a default (also fallback) model

PR #340, opened 2024-05-10
Expose model abstraction in frontend

yoziru:
Exposes model abstraction settings in frontend
Screenshot 2024-05-10 at 14 27 31
when using Azure, only the default settings are shown
Screenshot 2024-05-10 at 14 43 27

Split from https://github.com/grafana/grafana-llm-app/pull/331
Resolves https://github.com/grafana/grafana-llm-app/issues/323


sd2k (approved):
LGTM! Would like to remove the custom Azure deployments config eventually but that's a later thing.

PR #341, opened 2024-05-10
llm-frontend: make model an optional field

yoziru:
https://github.com/grafana/grafana-llm-app/pull/338 made this field optional, with a fallback to default (ModelBase)


packages/grafana-llm-frontend/src/openai.ts
```diff
@@ -90,7 +90,7 @@ export interface ChatCompletionsRequest {
   /**
    * Model abstraction to use. These abstractions are then translated back into specific models based on the users settings.
```
sd2k, line 91:
```suggestion
   * Model abstraction to use. These abstractions are then translated back into specific models based on the users settings.
   *
   * If not specified, defaults to `Model.BASE`.
```

sd2k (approved):
LGTM, just one doc suggestion

PR #342, opened 2024-05-10
Workaround openai issue with temperature: 0 being omitted from request

sd2k:
This PR adds a workaround for [this GitHub discussion][gh], describing
how chat completion requests which explicitly set temperature = 0 are
marshalled to JSON incorrectly (due to the 0 being indistinguishable
from the zero value of the field).

To do so we add a custom unmarshal method which checks whether the field
was defined and set to zero, and explicitly set the temperature to
`math.SmallestNonzeroFloat32` in such cases. This isn't perfect but is
likely to get very similar results in practice.

[gh]: https://github.com/sashabaranov/go-openai/issues/9#issuecomment-956843106

### DIFF
--- a/packages/grafana-llm-app/pkg/plugin/llm_provider.go
+++ b/packages/grafana-llm-app/pkg/plugin/llm_provider.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


yoziru (approved):
Close enough, works 👍

PR #349, opened 2024-05-13
Avoid using global variable to determine whether to ignore padding

### LINKED ISSUE #332: Flakey test due to padding switch
The `TestChatCompletionStreamResponseMarshalJSON` test is flakey, presumably because it relies on a global flag being switched on/off by certain tests which is non-deterministic depending on which tests run in what order.

To repro, run:

```sh
go test ./... -count 100 -run TestChatCompletionStreamResponseMarshalJSON
```

from the `packages/grafana-llm-app` directory.

sd2k:
Instead, use a non-configurable flag on the app which can only be set manually,
and set it to true in the test where we need to test response bodies.

Also, ensure we always send at least _some_ padding, since the padding field is tagged with `omitempty`.

Fixes #332.

### DIFF
--- a/packages/grafana-llm-app/pkg/plugin/app.go
+++ b/packages/grafana-llm-app/pkg/plugin/app.go
@@ -38,6 +38,10 @@ type App struct {
 	settings         *Settings
 	saToken          string
 	grafanaAppURL    string
+
+	// ignoreResponsePadding is a flag to ignore padding in responses.
+	// It should only ever be set in tests.
+	ignoreResponsePadding bool
 }
 
 // NewApp creates a new example *App instance.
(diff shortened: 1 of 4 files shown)


yoziru (approved):
```
❯ go test ./... -count 100 -run TestChatCompletionStreamResponseMarshalJSON

?   	github.com/grafana/grafana-llm-app/pkg	[no test files]
?   	github.com/grafana/grafana-llm-app/pkg/plugin/vector	[no test files]
?   	github.com/grafana/grafana-llm-app/pkg/plugin/vector/embed	[no test files]
?   	github.com/grafana/grafana-llm-app/pkg/plugin/vector/store	[no test files]
ok  	github.com/grafana/grafana-llm-app/pkg/plugin	0.294s
```

nice one ✅

PR #350, opened 2024-05-13
Differentiate between disabled and not configured

csmarchbanks:
In order to support UX flows to prompt users to configure the LLM plugin if they have not we need to differentiate between a state where an admin has actively disabled LLM functionality and just not configured any LLM functionality. Do this by adding a `disabled` flag to the OpenAI configuration which will be set to true when an admin chooses the disable option. By default, no option will be chosen.

The health endpoint's `configured` field can be used to determine if the app needs to be configured (not disabled, and incomplete configuration).


sd2k (approved):
LGTM, tested and does what I expected :+1:

PR #378, opened 2024-06-12
fix: remove reference to deprecated CSS class

### LINKED ISSUE #377: Replace deprecated Grafana SCSS styles
Hi from the Grafana team 👋 

In Grafana 12, we'll be [removing Angular support](https://grafana.com/blog/2024/03/11/removal-of-angularjs-support-in-grafana-what-you-need-to-know/). As part of that, we'll also be removing a lot of legacy CSS classes.
It looks like this plugin relies on one of these classes:
- `p-t-2`

The CSS class names that Grafana provides can change at any time and should not be relied on as a stable external API. The recommendation would be to either:
- [Use the theme object with emotion](https://github.com/grafana/grafana/blob/main/contribute/style-guides/styling.md) to style your components. The theme object is a stable API that is maintained according to SemVer. 
- Apply these styles yourself within this repo. You can find the existing styles [here](https://github.com/grafana/grafana/blob/main/public/sass/_angular.scss) as a reference.

Thanks again for your contributions and support! 🙌

sd2k:
For some reason we were using a div with gf-form-group
and p-t-2 classes, which feels unnecessary. I've just removed
the wrapper div because I don't think it does anything
(it probably came from a copy/paste).

Fixes #377.

### DIFF
--- a/packages/grafana-llm-app/src/components/AppConfig/HealthCheck.tsx
+++ b/packages/grafana-llm-app/src/components/AppConfig/HealthCheck.tsx
@@ -71,11 +71,7 @@ const getAlertSeverity = (status: string, details: HealthCheckDetails): AlertVar
 export function ShowHealthCheckResult(props: HealthCheckResult) {
   let severity = getAlertVariant(props.status ?? 'error');
   if (!isHealthCheckDetails(props.details)) {
-    return (
-      <div className="gf-form-group p-t-2">
-        <Alert severity={severity} title={props.message} />
-      </div>
-    );
+    return <Alert severity={severity} title={props.message} />;
   }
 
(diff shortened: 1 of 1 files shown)


ashharrison90 (approved):
that was quick! 🚀 

- `gf-form-group` effectively sets `marginBottom: theme.spacing(5)` (see https://github.com/grafana/grafana/blob/main/public/sass/components/_gf-form.scss#L77)
- `p-t-2` effectively sets `paddingTop: theme.spacing(3)`

i have no idea where this component appears or whether you still need those styles. if the `<Alert>` is appearing inline you might still need them. if you do, you could just apply them as normal emotion styles 👍

PR #404, opened 2024-07-17
Add validation to OpenAI URL

yoziru:
Invalidate URLs that have spaces in the hostname since golang doesn't like that
e.g. `https://api.open ai.com` would throw an error `invalid character "" in host name`

This prevents saving invalid URLs in the form:
Screenshot 2024-07-17 at 13 47 40

### DIFF
--- a/packages/grafana-llm-app/pkg/plugin/openai_provider.go
+++ b/packages/grafana-llm-app/pkg/plugin/openai_provider.go
@@ -24,6 +24,11 @@ func NewOpenAIProvider(settings OpenAISettings, models *ModelSettings) (LLMProvi
 		Timeout: 2 * time.Minute,
 	}
 	cfg := openai.DefaultConfig(settings.apiKey)
+	// validate url
+	_, err := url.Parse(settings.URL)
+	if err != nil {
+		return nil, fmt.Errorf("invalid url: %w", err)
+	}
 	base, err := url.JoinPath(settings.URL, "/v1")
 	if err != nil {
 		return nil, fmt.Errorf("join url: %w", err)
--- a/packages/grafana-llm-app/src/components/AppConfig/AppConfig.tsx
+++ b/packages/grafana-llm-app/src/components/AppConfig/AppConfig.tsx
@@ -14,6 +14,23 @@ import { OpenAISettings } from './OpenAI';
 import { VectorConfig, VectorSettings } from './Vector';
 ///////////////////////
 
+const validateUrl = (url: string): [boolean, string] => {
+  try {
+    const parsedURL = new URL(url);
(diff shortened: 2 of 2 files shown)


packages/grafana-llm-app/pkg/plugin/openai_provider.go
```diff
@@ -24,6 +24,11 @@ func NewOpenAIProvider(settings OpenAISettings, models *ModelSettings) (LLMProvi
 		Timeout: 2 * time.Minute,
 	}
 	cfg := openai.DefaultConfig(settings.apiKey)
+	// validate url
+	_, err := url.Parse(settings.URL)
+	if err != nil {
+		return nil, fmt.Errorf("invalid url: %w", err)
```
csmarchbanks, line 30: Is a hard error the correct thing to do here? I think that means for any plugins with current bad data, or that are provisioned incorrectly then the plugin will fail to load which would also mean the setting cannot be updated? Could we have a soft failure instead?

gerboland:
I've added a PR that I think might do.

It appears to be a convention that Grafana settings pages let you input invalid details and save them, only exposing the problem when the health-check fails later ("Save & test" button). The PR maintains that convention, but invalid settings can now be fixed.

packages/grafana-llm-app/src/components/AppConfig/AppConfig.tsx
```diff
@@ -14,6 +14,23 @@ import { OpenAISettings } from './OpenAI';
 import { VectorConfig, VectorSettings } from './Vector';
 ///////////////////////
 
+const validateUrl = (url: string): [boolean, string] => {
```
gerboland, line 17: Why not `url.ParseRequestURI` ?

gerboland, line 17: `URL.canParse` might be better than something hand-rolled?

yoziru: Makes sense, still need some custom rules to handle diffs between go validation and js validation
e.g. JS
```js
URL.canParse("http://api.open ai.com")
> true
```
Golang
```go
u, err := url.Parse("http://api.open ai.com")
> panic: parse "http://api.open ai.com": invalid character " " in host name
```

gerboland:
```javascript
URL.canParse("http://api.open ai.com")
> true
```

geez...

yoziru:
Closing since https://github.com/grafana/grafana-llm-app/pull/405 fails the health check and serves the same purpose in a better way

PR #405, opened 2024-07-17
Create LLMProvider only when needed

gerboland:
When reading/writing settings, we don't need to create any LLMProvider.

Avoiding creating the LLMProvider means we avoid validating the settings passed to the provider, which may be incorrect.

Fixes https://github.com/grafana/support-escalations/issues/11496


csmarchbanks (approved):
I like this approach, my only thought is that is that it might be nice to cache the provider for the streaming case as well, but not a big deal as creating a provider should be pretty quick.

yoziru (approved):

Issue #421, opened 2024-08-16
llm-app: Support for Assistant APIs?


paulcoghlan:
It would be great if the [llmclient](https://github.com/grafana/grafana-llm-app/blob/main/packages/grafana-llm-app/llmclient/llmclient.go) could support more endpoints to enable the assistant features, namely:

```
        CreateAssistant(ctx context.Context, request openai.AssistantRequest) (openai.Assistant, error)
	CreateThread(ctx context.Context, request openai.ThreadRequest) (openai.Thread, error)
	CreateMessage(ctx context.Context, threadID string, request openai.MessageRequest) (msg openai.Message, err error)
	CreateRun(ctx context.Context, threadID string, request openai.RunRequest) (run openai.Run, err error)
	RetrieveRun(ctx context.Context, threadID string, runID string) (run openai.Run, err error)
	SubmitToolOutputs(ctx context.Context, threadID string, runID string, request openai.SubmitToolOutputsRequest) (response openai.Run, err error)
	ListMessage(ctx context.Context, threadID string) (openai.MessagesList, error)
```

Happy to help with a PR if necessary.

gerboland:
Would be worth following up with the team to see what they've done since. Are they using OpenAI's api's directly? If so, we should try include the functionality in the llm-app...

jtheory:
( following up in Slack with @paulcoghlan )

PR #464, opened 2024-10-07
Add support for Assistant APIs

paulcoghlan:
Relates to: https://github.com/grafana/grafana-llm-app/issues/421

It would be great if the LLM plugin could support some Assistant features, so I've raised this PR to add support for `Assistant`s, `Thread`s, `Run`s and `Message`s.

I've removed the `go.mod` file from the `llmclient` so we support the latest version of http://github.com/sashabaranov/go-openai.

Any feedback welcome as it is my first PR! 🙏 

## Health Check Scenarios

1. Models OK, Assistant Supported
Screenshot 2024-11-04 at 19 09 32

```
GET /api/plugins/grafana-llm-app/health
{
    "details": {
        "openAI": {
            "assistant": {
                "ok": true
            },
            "configured": true,
            "models": {
                "base": {
                    "ok": true
                },
                "large": {
                    "ok": true
                }
            },
            "ok": true
        },
   ...
      "version": "0.11.0"
    },
    "message": "",
    "status": "OK"
}
```

2. Models OK, Assistant Not Supported
```
GET /api/plugins/grafana-llm-app/health
{
    "details": {
        "openAI": {
            "assistant": {
                "ok": fail,
                "error": "Assistant not present"
            },
            "configured": true,
            "models": {
                "base": {
                    "ok": true
                },
                "large": {
                    "ok": true
                }
            },
            "ok": true
        },
   ...
      "version": "0.11.0"
    },
    "message": "",
    "status": "OK"
}
```

3. Models Fail
As previously - failure

### DIFF
--- a/packages/grafana-llm-app/llmclient/go.mod
+++ b/packages/grafana-llm-app/llmclient/go.mod
@@ -1,5 +1,0 @@
-module github.com/grafana/grafana-llm-app/llmclient
-
-go 1.19
-
-require github.com/sashabaranov/go-openai v1.15.3
--- a/packages/grafana-llm-app/llmclient/go.mod
+++ b/packages/grafana-llm-app/llmclient/go.mod
@@ -2,4 +2,4 @@ module github.com/grafana/grafana-llm-app/llmclient
 
 go 1.19
 
-require github.com/sashabaranov/go-openai v1.15.3
+require github.com/sashabaranov/go-openai v1.32.0
--- a/packages/grafana-llm-app/llmclient/go.mod
+++ b/packages/grafana-llm-app/llmclient/go.mod
@@ -1,0 +1,5 @@
+module github.com/grafana/grafana-llm-app/llmclient
+
+go 1.19
+
+require github.com/sashabaranov/go-openai v1.15.3
--- a/packages/grafana-llm-app/llmclient/llmclient.go
+++ b/packages/grafana-llm-app/llmclient/llmclient.go
@@ -55,14 +55,32 @@ type OpenAI interface {
 	ChatCompletionsStream(ctx context.Context, req ChatCompletionRequest) (*openai.ChatCompletionStream, error)
 	// CreateAssistant creates an assistant using the given request.
 	CreateAssistant(ctx context.Context, req AssistantRequest) (openai.Assistant, error)
+	// RetrieveAssistant retrieves an assistant by ID.
+	RetrieveAssistant(ctx context.Context, assistantID string) (openai.Assistant, error)
+	// ListAssistants lists assistants.
+	ListAssistants(ctx context.Context, limit *int, order *string, after *string, before *string) (openai.AssistantsList, error)
+	// DeleteAssistant deletes an assistant by ID.
+	DeleteAssistant(ctx context.Context, assistantID string) (openai.AssistantDeleteResponse, error)
 	// CreateThread creates a new thread.
 	CreateThread(ctx context.Context, req openai.ThreadRequest) (openai.Thread, error)
+	// RetrieveThread retrieves a thread by ID.
+	RetrieveThread(ctx context.Context, threadID string) (openai.Thread, error)
+	// DeleteThread deletes a thread by ID.
+	DeleteThread(ctx context.Context, threadID string) (openai.ThreadDeleteResponse, error)
 	// CreateMessage creates a new message in a thread.
 	CreateMessage(ctx context.Context, threadID string, request openai.MessageRequest) (msg openai.Message, err error)
+	// ListMessages lists messages in a thread.
+	ListMessages(ctx context.Context, threadID string, limit *int, order *string, after *string, before *string) (openai.MessagesList, error)
+	// RetrieveMessage retrieves a message in a thread.
+	RetrieveMessage(ctx context.Context, threadID string, messageID string) (msg openai.Message, err error)
+	// DeleteMessage deletes a message in a thread.
+	DeleteMessage(ctx context.Context, threadID string, messageID string) (msg openai.MessageDeletionStatus, err error)
 	// CreateRun creates a new run in a thread.
 	CreateRun(ctx context.Context, threadID string, request openai.RunRequest) (run openai.Run, err error)
 	// RetrieveRun retrieves a run in a thread.
 	RetrieveRun(ctx context.Context, threadID string, runID string) (run openai.Run, err error)
+	// CancelRun cancels a run in a thread.
+	CancelRun(ctx context.Context, threadID string, runID string) (run openai.Run, err error)
 	// SubmitToolOutputs submits tool outputs for a run in a thread.
 	SubmitToolOutputs(ctx context.Context, threadID string, runID string, request openai.SubmitToolOutputsRequest) (response openai.Run, err error)
 	// ListMessage lists messages in a thread.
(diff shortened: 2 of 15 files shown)


sd2k:
Thanks Paul! In general this looks good, our main concern is how to handle LLM providers which don't support these APIs (e.g. if the user has configured a custom OpenAI-compatible LLM backend other than OpenAI). We may need to add something to the health check response which checks whether these things exist, so that clients can degrade gracefully in those cases.

sd2k:
Yep sounds like a plan. It could maybe go inside the `OpenAIHealthDetails` type though since it's related to the level of OpenAI support :+1:

paulcoghlan:
> Yep sounds like a plan. It could maybe go inside the `OpenAIHealthDetails` type though since it's related to the level of OpenAI support 👍

Got it - thanks, will do 😄.

packages/grafana-llm-app/pkg/plugin/health.go
```diff
@@ -86,6 +98,7 @@ func (a *App) openAIHealth(ctx context.Context, req *backend.CheckHealthRequest)
 		OK:         true,
 		Configured: a.settings.OpenAI.Configured(),
 		Models:     map[Model]openAIModelHealth{},
+		Assistant:  openAIModelHealth{OK: false, Error: "Assistant not present"},
```
sd2k, line 101:
```suggestion
		Assistant:  openAIModelHealth{OK: false, Error: "Assistant not available"},
```

I kinda like `available` more but this is very much personal preference!

packages/grafana-llm-app/pkg/plugin/health.go
```diff
@@ -113,6 +126,14 @@ func (a *App) openAIHealth(ctx context.Context, req *backend.CheckHealthRequest)
 		d.Error = "No functioning models are available"
 	}
 
+	if d.Configured {
+		err := a.testOpenAIAssistant(ctx)
+		if err == nil {
+			d.Assistant.OK = true
+			d.Assistant.Error = ""
+		}
```
sd2k, line 134: Perhaps we should append the error to the `d.Assistant.Error` if it's non-nil, to make it easier for users to debug why the assistant isn't available.

packages/grafana-llm-app/pkg/plugin/llm_provider.go
```diff
@@ -133,4 +133,6 @@ type LLMProvider interface {
 	// ChatCompletionStream provides text completion in a chat-like interface with
 	// tokens being sent as they are ready.
 	ChatCompletionStream(context.Context, ChatCompletionRequest) (<-chan ChatCompletionStreamResponse, error)
+	// ListAssistants lists assistants.
```
sd2k, line 136:
```suggestion
	// ListAssistants lists assistants.
	//
	// This is used by the health check to determine whether the configured provider
	// supports assistant APIs.
```

packages/grafana-llm-app/src/components/AppConfig/HealthCheck.tsx
```diff
@@ -128,6 +137,8 @@ function ShowOpenAIHealth({ openAI }: { openAI: OpenAIHealthDetails | boolean })
           </li>
         ))}
       </div>
+      <b>Assistant: </b>
+      {openAI.assistant.ok ? 'OK' : `Error: ${openAI.assistant.error}`}
```
sd2k, line 141:
```suggestion
      {openAI.assistant.ok ? 'OK' : `Error: ${openAI.assistant.error}. The configured OpenAI provider may not offer assistants APIs.`}
```

packages/grafana-llm-frontend/src/types.ts
```diff
@@ -21,6 +21,8 @@ export interface OpenAIHealthDetails {
   // The health check attempts to call the OpenAI API with each
   // of a few models and records the result of each call here.
   models?: Record<string, OpenAIModelHealthDetails>;
+  // Health details for the OpenAI assistant model.
```
sd2k, line 24:
```suggestion
  // Health details for the OpenAI assistant APIs.
```

sd2k (commented):
This is looking good to me, thanks Paul! I've added a few suggestions inline but would like to get a second opinion on a few things, mostly:

- the go.mod change (I think this should at least be in a separate PR, and need to learn more about it first)
- the now-large `OpenAI` interface in `llmclient`; perhaps it's worth splitting it

@csmarchbanks could you take a look at this when you have time?

packages/grafana-llm-app/llmclient/llmclient.go
```diff
@@ -47,6 +53,36 @@ type OpenAI interface {
...
+	// DeleteMessage deletes a message in a thread.
+	DeleteMessage(ctx context.Context, threadID string, messageID string) (msg openai.MessageDeletionStatus, err error)
+	// CreateRun creates a new run in a thread.
+	CreateRun(ctx context.Context, threadID string, request openai.RunRequest) (run openai.Run, err error)
+	// RetrieveRun retrieves a run in a thread.
+	RetrieveRun(ctx context.Context, threadID string, runID string) (run openai.Run, err error)
+	// CancelRun cancels a run in a thread.
+	CancelRun(ctx context.Context, threadID string, runID string) (run openai.Run, err error)
+	// SubmitToolOutputs submits tool outputs for a run in a thread.
+	SubmitToolOutputs(ctx context.Context, threadID string, runID string, request openai.SubmitToolOutputsRequest) (response openai.Run, err error)
```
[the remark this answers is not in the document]
csmarchbanks: I also think having a second interface could be nice. It would also allow us to check for features based on if the interface is implemented for a connection or not (in the case we add more first class implementations of this interface).

packages/grafana-llm-app/llmclient/go.mod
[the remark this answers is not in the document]
csmarchbanks: Yeah, what's the reason for this removal? I like specifying that we depend on >=1.15.3, <2.0.0. Are you running into import issues somewhere?

packages/grafana-llm-app/pkg/plugin/health.go
paulcoghlan: Good point, will do.

packages/grafana-llm-app/src/components/AppConfig/HealthCheck.tsx
paulcoghlan: Agree, this is a more precise description.

packages/grafana-llm-app/llmclient/llmclient.go
```diff
@@ -63,9 +108,14 @@ func NewOpenAI(grafanaURL, grafanaAPIKey string) OpenAI {
...
 
+func NewOpenAIAssistant(grafanaURL, grafanaAPIKey string) OpenAIAssistant {
+	httpClient := &http.Client{}
+	return NewOpenAIWithClient(grafanaURL, grafanaAPIKey, httpClient)
+}
+
 // NewOpenAIWithClient creates a new OpenAI client talking to the Grafana LLM app installed
 // on the given Grafana instance, using the given HTTP client.
-func NewOpenAIWithClient(grafanaURL, grafanaAPIKey string, httpClient *http.Client) OpenAI {
+func NewOpenAIWithClient(grafanaURL, grafanaAPIKey string, httpClient *http.Client) *openAI {
```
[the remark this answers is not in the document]
paulcoghlan: Oops, that was temporary change whilst I was testing something out! Will revert.

packages/grafana-llm-app/llmclient/llmclient.go
```diff
@@ -80,6 +132,12 @@ func NewOpenAIWithClient(grafanaURL, grafanaAPIKey string, httpClient *http.Clie
 	}
 }
 
+// NewOpenAIAssistantWithClient creates a new OpenAI client talking to the Grafana LLM app installed
+// on the given Grafana instance, using the given HTTP client.
+func NewOpenAIAssistantWithClient(grafanaURL, grafanaAPIKey string, httpClient *http.Client) OpenAIAssistant {
+	return NewOpenAIWithClient(grafanaURL, grafanaAPIKey, httpClient).(OpenAIAssistant)
```
sd2k, line 138: nice! 👨‍🍳 💋

sd2k (approved):
This looks good to me now! Approving but it might be worth waiting for @csmarchbanks to take a look too.

SandersAaronD (approved):
These look good to me, will merge.

Worth noting: It makes perfect sense to make sure we have this functionality, but unlike the general openai endpoint, which has become very very standard, this is both vendor specific and in beta currently. We should try to make sure we do as much as we can to make sure it's as modular as possible and doesn't interfere with using other vendors going forward, assuming it does not also become a nigh-ubiquitous standard.

PR #471, opened 2024-10-18
bump grafana-llm-app to 0.11.0

yoziru:


sd2k (commented):
Do we want to do a minor bump since the models have changed or is it not a big deal?

yoziru:
> Do we want to do a minor bump since the models have changed or is it not a big deal?

Not sure, there is no functional or API change apart from better completions
Screenshot 2024-10-18 at 13 44 30

But happy to do a minor bump, feels safer

jtheory (approved):
LGTM (after versioning discussion)

PR #478, opened 2024-11-04
chore: relax version specifiers in frontend library

sd2k:
We're a library, so we shouldn't specify exact versions of things,
or downstream users may end up with multiple versions of a dependency
(e.g. React).


gitdoluquita (approved):
Very relaxed 😌

PR #500, opened 2024-12-10
Update dependencies to remediate various security warnings

SandersAaronD:
This should moot a few of the dependabot updates, which tried to update these libraries one at a time. Updated in sync they seem to not have any breakages on this repo.


SandersAaronD:
This is going to fail e2e tests, cypress is no longer recommended (or updated!) and breaks with the (important, security) package updates. Cypress tests aren't actually used anyway here, so they're removed, but it does mean that it will need a click-through to bypass CI here.

Will cut an issue to migrate to playwright.

SandersAaronD:
Updated the rest of dep bumps except the go-openai one to save us some testing (in case any of them behave differently from the new main).

sd2k (approved):
LGTM :+1:

PR #557, opened 2025-02-06
Configs: Add custom API as an separate option

edwardcqian:
Currently the plugin supports custom API. However, the UI is not very clear.

Extract the "OpenAI compatible" setup logic to it's own "Custom API" card.

uses "OpenAI-like" to describe the API signature to make it more clear that it's not OpenAI

### DIFF
--- a/packages/grafana-llm-app/pkg/plugin/provider.go
+++ b/packages/grafana-llm-app/pkg/plugin/provider.go
@@ -6,6 +6,8 @@ func createProvider(settings *Settings) (LLMProvider, error) {
 	switch settings.OpenAI.Provider {
 	case openAIProviderOpenAI:
 		return NewOpenAIProvider(settings.OpenAI, settings.Models)
+	case openAIProviderCustom:
+		return NewOpenAIProvider(settings.OpenAI, settings.Models)
 	case openAIProviderAzure:
 		return NewAzureProvider(settings.OpenAI, settings.Models.Default)
 	case openAIProviderGrafana:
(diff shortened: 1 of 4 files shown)


packages/grafana-llm-app/pkg/plugin/provider.go
```diff
@@ -6,6 +6,8 @@ func createProvider(settings *Settings) (LLMProvider, error) {
 	switch settings.OpenAI.Provider {
 	case openAIProviderOpenAI:
```
sd2k, line 7: nit: you can do `case openAIProviderOpenAI, openAIProviderCustom:` here if you like!

packages/grafana-llm-app/src/components/AppConfig/OpenAI.tsx
```diff
@@ -45,22 +45,35 @@ export function OpenAIConfig({
         event.currentTarget.type === 'checkbox' ? event.currentTarget.checked : event.currentTarget.value.trim(),
     });
   };
+
+  // Update settings when provider changes, set default URL for OpenAI
+  const onChangeProvider = (value: OpenAIProvider) => {
+    onChange({
+      ...settings,
+      provider: value,
+      url: value === 'openai' ? 'https://api.openai.com' : '',
```
sd2k, line 54: nit: might be worth extracting this URL into a constant since we use it in a few places

sd2k (approved):
LGTM but will leave it up for more UX-minded focus to check wording :sweat_smile:

csmarchbanks (approved):

Issue #504, opened 2024-12-11
Add support for Anthropic API


annanay25:
Recently we've seen great performance from the Claude model family and it would be nice to add support for these models to the LLM app. 

API reference: https://docs.anthropic.com/en/api/messages
Typescript SDK: https://github.com/anthropics/anthropic-sdk-typescript

csmarchbanks:
Let's start with a spike for determining pros/cons of doing this in openrouter vs implementing directly in the plugin backend.

sd2k:
For doing it directly: ideally we'd be able to find something like the [litellm](https://docs.litellm.ai/docs/) Python library which lets you call tons of LLM providers using an OpenAI-like request/response format, but I can't seem to find one for Go; we'd probably need to re-implement the [Anthropic part of litellm](https://github.com/BerriAI/litellm/blob/main/litellm/llms/anthropic/chat/transformation.py#L481-L695) ourselves and keep it maintained.

ishaan-jaff:
Hi @sd2k thanks for considering litellm. If you'd like you can also use our proxy server https://docs.litellm.ai/docs/proxy/docker_quick_start and just point to it 

That way you don't need to maintain code in GO - litellm maintainer

PR #561, opened 2025-02-13
Abstract provider configuration

gitdoluquita:
This PR implements the foundational changes needed to support the Anthropic API integration (#504). It refactors the provider configuration infrastructure to handle multiple LLM providers in a more generic way.

### Key Changes
- Abstracts provider configuration by moving provider type to the top-level Settings struct
- Relocated `disabled` flag from OpenAI settings to top-level settings to make it provider-agnostic
- Adds initial AnthropicSettings struct with basic configuration fields (URL, API key)
- Maintains backward compatibility for existing OpenAI configurations

### Next Steps
-  Implement Anthropic provider client
- Add UI components for Anthropic configuration
- Add support for Anthropic model mappings
- Implement message streaming for Anthropic API
- Add tests for Anthropic provider

The changes maintain existing functionality while preparing for Anthropic support. All existing tests pass with the refactored provider configuration.

#504 (partially - this is the first step)

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Each package has its own package.json with its own npm commands, but the intenti
 3. Run `npm install`
 4. Run `npm run dev`
 5. In a separate terminal from dev, run `npm run server`
-If you want to bring up the vector services, you can run `COMPOSE_PROFILES=vector yarn run server` instead
+If you want to bring up the vector services, you can run `COMPOSE_PROFILES=vector npm run server` instead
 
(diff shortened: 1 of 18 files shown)


sd2k (commented):
This LGTM, nice and comprehensive, thanks!

Only one request, and it's not a blocker, but could we separate out the addition of the Anthropic settings into a separate PR? Since we're adding some migration stuff to this (which _may_ have to be reverted, although hopefully not), it would be nice to have it separate.

sd2k (approved):
Tested and works for me :+1: Nice!

PR #565, opened 2025-02-19
refactor: rename OpenAI to LLMProvider for vendor neutrality

gitdoluquita:
This PR refactors the codebase to use vendor-neutral terminology instead of OpenAI-specific naming. This is part of issue #504 to add support for Anthropic and other LLM providers.

## Key Changes
- Renamed the OpenAI interface to LLMProvider to better reflect its vendor-neutral role
- Added new /llm/v1/ API endpoints that will serve as the standard endpoint path for all LLM providers
- Deprecated existing /openai/v1/ endpoints but maintained backwards compatibility
- Created a new vendor-neutral llm.ts module and deprecated the OpenAI-specific openai.ts
- Updated health check types and UI components to use provider-neutral terminology
- Updated tests to reflect the new naming conventions


edwardcqian (approved):
test and everything still working as expected

PR #566, opened 2025-02-19
Add Anthropic Provider Support

### LINKED ISSUE #504: Add support for Anthropic API

gitdoluquita:
This PR adds support for using Anthropic's Claude models as an LLM provider in the Grafana LLM app. Users can now choose between OpenAI, Anthropic, and other providers when configuring LLM features.

## Key changes

- Added new `ProviderTypeAnthropic` and associated settings structures
- Integrated Anthropic SDK and default model mappings
- Updated UI to include Anthropic as a provider option with dedicated configuration panel
- Updated provider factory to handle Anthropic provider type

Closes #504

### DIFF
--- a/packages/grafana-llm-app/README.md
+++ b/packages/grafana-llm-app/README.md
@@ -69,8 +69,8 @@ apps:
   - type: 'grafana-llm-app'
     disabled: false
     jsonData:
+      provider: azure
       openAI:
-        provider: azure
         url: https://<resource>.openai.azure.com
         azureModelMapping:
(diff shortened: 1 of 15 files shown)


gerboland (approved):
Tested locally, appears to work just fine!

PR #570, opened 2025-02-24
tweak health check assistant message

yoziru:
- **replace deprecated VerticalGroup with Stack**
- **separate optional assistant api health check message**

Assistant API is optional / internal use for now, so just add message in separate info alert
Was thinking we could also just hide it entirely (except show it if the check is OK)?
Screenshot 2025-02-24 at 15 36 22


gitdoluquita (approved):
LGTM!

PR #574, opened 2025-02-27
feat: run MCP server in backend using Grafana Live

sd2k:
This plugin starts up an MCP server using a custom transport based on
Grafana Live. This is similar to the SSE transport, but rather than
using a long-lived SSE connection to return events, the MCP server
returns them on a Grafana Live subscription. Similarly, rather than
accepting JSON-RPC messages on the `/messages` endpoint, we accept
messages over the `PublishStream` Grafana Live handler. This gets us
similar functionality to the SSE transport, including all those which
require statefulness.

The tools are imported directly from the [Grafana MCP repository][mcp-grafana]
so will be up-to-date with the latest version of the MCP server.

The original PR included some additional changes to the frontend to make
it easier for users to use the MCP functionality. These changes have
been split out into a separate PR to make this one easier to review.

[mcp-grafana]: https://github.com/grafana/mcp-grafana

---

**note: these instructions are out of date for now, I'll update soon**

There's an example of this (and some frontend updates) being used in [this PR](https://github.com/grafana/grafana-plugin-examples/pull/480) to the `grafana-plugin-examples` repo. The simplest way to test it out is probably to:

- build the plugin and frontend library in this PR (`npm i && run build:all`)
- clone the `grafana-plugin-examples` repo alongside this one
- cd into the `grafana-plugin-examples/examples/app-with-mcp` directory
- make sure the relative directories in docker-compose.yaml and package.json are correct for your filesystem
- run `mage`, `npm i`, `npm run dev` to build the sample app
- run `docker compose up` to start the sample app's Grafana
- head to http://localhost:3000/a/grafana-appwithmcp-app/one to see the example page


sd2k:
Marking this as ready even though it's not mergeable yet, would love opinions on the approach and frontend APIs :slightly_smiling_face:

sd2k:
One thing I've noticed doing this is that our OpenAI types in the frontend library are waaay out of date. The [official OpenAI JS/TS library](https://www.npmjs.com/package/openai?activeTab=readme) is focused on node though so it feels a bit wrong to pull it all in just for type definitions. We should try and find a better solution than what we currently have.

packages/grafana-llm-app/pkg/plugin/app.go
```diff
@@ -105,4 +113,18 @@ func (a *App) Dispose() {
 	if a.vectorService != nil {
 		a.vectorService.Cancel()
 	}
+	if a.mcpServer != nil {
+		a.mcpServer.Close()
+	}
+}
+
+func newMcpServer() *mcp.GrafanaLiveServer {
+	// TODO: get correct version somehow. Build info?
```
csmarchbanks, line 122: What version are you trying to get here? The version of the llm-app or of the MCP server embedded in the llm app?  If the former, ldflags are the way to go, if the latter you could do something like:
```go
version := "unknown"
buildInfo, ok := debug.ReadBuildInfo()
if ok {
	for _, dep := range buildInfo.Deps {
		if dep.Path == "github.com/grafana/mcp-grafana" {
			version = dep.Version
		}
	}
}
```

csmarchbanks (approved):
I am happy to have this in despite the couple TODOs that we can iterate on. It works with the frontend updates in the other PR, and we can keep improving it.

packages/grafana-llm-app/pkg/mcp/live_server.go
```diff
@@ -0,0 +1,221 @@
...
+	if url.Scheme == "http" {
+		t.Schemes = []string{"http"}
+	}
+
+	// TODO: fetch ID token / auth token from headers.
+
+	// Hmm, we have the user here, but no ID token? How do we make requests as the user?
+	// req.PluginContext.User
+	// This will authenticate as the plugin not the current user, which isn't what we want,
+	// as the plugin may have more permissions than the user.
```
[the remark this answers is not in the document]
sd2k: Yep good call! Thanks.

packages/grafana-llm-app/pkg/plugin/app.go
sd2k: I think we need the LLM app version here. I've used linker flags in the most recent commit. They work because the `mage` command we use to build sets them:

```sh
$ mage -v
exec: go "build" "-o" "dist/gpx_llm_linux_amd64" "-tags" "arrow_json_stdlib" "-ldflags" "-w -s -extldflags \"-static\" -X 'github.com/grafana/grafana-plugin-sdk-go/build.buildInfoJSON={\"time\":1741880085700,\"pluginID\":\"grafana-llm-app\",\"version\":\"0.13.2\"}' -X 'main.pluginID=grafana-llm-app' -X 'main.version=0.13.2'" "./pkg"
```

packages/grafana-llm-app/pkg/mcp/live_server.go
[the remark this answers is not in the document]
sd2k: Hmm since we're using the app's service account token, I think we'll need to update our plugin's permissions in plugin.json for the service account to have access to e.g. list datasources.

packages/grafana-llm-app/pkg/main.go
```diff
@@ -8,8 +8,14 @@ import (
...
 )
 
+// version is the version of the plugin.
+// It is set at build time by the Mage target, which passes
+// the version to the Go linker using `-X main.version=x.y.z`.
+var version = "development"
+
 func main() {
-	log.DefaultLogger.Info("Starting plugin process")
+	plugin.PluginVersion = version
```
csmarchbanks, line 17: should this be done in an `init()` next to the variable definition instead?

csmarchbanks: I guess we can't really due to the dependency order. I guess this is fine, though maybe `init()` is still more correct than `main()`.

sd2k: Yup that's better, thanks.

packages/grafana-llm-app/pkg/mcp/live_server.go
```diff
@@ -0,0 +1,218 @@
...
+	cfg := backend.GrafanaConfigFromContext(ctx)
+	if cfg == nil {
+		return ctx
+	}
+	grafanaURL, err := cfg.AppURL()
+	if err != nil {
+		return ctx
+	}
+	apiKey, _ := cfg.PluginAppClientSecret()
+	incidentUrl := fmt.Sprintf("%s/api/plugins/grafana-incident-app/resources/api/", strings.TrimSuffix(grafanaURL, "/"))
```
sd2k, line 199: We may need to update this to `grafana-irm-app` at some point but it's OK for now.

csmarchbanks (approved):
Seems like a reasonable list of permissions to start with.

packages/grafana-llm-app/src/plugin.json
```diff
@@ -42,6 +42,30 @@
     "permissions": [
       {
         "action": "orgs:read"
+      },
+      {
+        "action": "datasources:read",
+        "scope": "datasources:uid:*"
```
csmarchbanks, line 48: Do we need the `:uid` here or just `datasources:*`? I see the latter on https://grafana.com/docs/grafana/latest/developers/http_api/data_source/.

sd2k: Good question, I figured both would work but have updated to match the docs.

packages/grafana-llm-app/src/plugin.json
```diff
@@ -42,6 +42,30 @@
...
+        "action": "folders:read",
+        "scope": "folders:*"
+      },
+      {
+        "action": "plugins.app:access",
+        "scope": "plugins:grafana-incident-app"
+      },
+      {
+        "action": "plugins.app:access",
+        "scope": "plugins:grafana-irm-app"
```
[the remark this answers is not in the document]
csmarchbanks: Yes they would. Ben and I had a quick offline discussion and decided to put the MCP feature behind a flag until on-behalf-of authentication is implemented. There could be too much that a user without appropriate permissions could use until then.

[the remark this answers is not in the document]
sd2k: I believe so, yeah. We need write access to create incidents, too.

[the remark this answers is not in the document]
annanay25: > Yes they would. Ben and I had a quick offline discussion and decided to put the MCP feature behind a flag until on-behalf-of authentication is implemented.

👍  We should document this on the LLM app docs so users know the ramifications before enabling it

annanay25 (approved):
lgtm!

PR #585, opened 2025-03-06
Bug: duplicate healthcheck details for backwards compatiblity

edwardcqian:
<0.13.0 grafana/llm frontend package check explicitly for openAI in the healthcheck.

Releasing plugin backend > 0.13.0 would cause all LLM powered features to stop working if they did not also have grafana/llm version > 0.13.0. Since changes in Grafana rolls out slowly this would require us to pin the plugin versions for certain Grafana versions and it would break things for OSS/enterprise.

This PR simply duplicates the llmProvider healthcheck details under openAI

**Consideration for reviewers:**

Alternatively, we could revert the healthcheck field llmProvider -> openAI, change frontend to check for llmProvider or openAI, and then roll rename later. Since the healthcheck detail/endpoint isn't very public, it might not be quite that bad.

### DIFF
--- a/packages/grafana-llm-app/pkg/plugin/health.go
+++ b/packages/grafana-llm-app/pkg/plugin/health.go
@@ -34,8 +34,10 @@ type vectorHealthDetails struct {
 
 type healthCheckDetails struct {
 	LLMProvider llmProviderHealthDetails `json:"llmProvider"`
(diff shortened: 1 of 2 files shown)


csmarchbanks:
I pushed a test as we ran into this in hackathon as well and will have issues if 0.13.x is not deployed.

sd2k (approved):

PR #590, opened 2025-03-12
 feat: add MCP helper functions to the `@grafana/llm` package

sd2k:
This commit adds some convenience functions to the `@grafana/llm`
Typescript package to make it possible for users to use the
MCP server now included in the LLM app.

Specifically it adds:

- a new transport implementation for the Grafana Live transport
  exposed by the LLM app's MCP server
- a helper function to convert MCP tools to OpenAI tools
- a Suspense-compatible React Provider to make it easy to get an
  initialized MCP client (otherwise quite annoying, since this is async)
- an example of how to do this, in the dev sandbox available in the
  LLM app's configuration page

Note that this is based on #574.

To test, you can just start up the dev process as usual and go to the config page. The dev sandbox will use tools if you use the 'Submit request' button (I haven't implemented it for the stream endpoint because it's quite complicated - will add some helpers for that in future).


packages/grafana-llm-frontend/src/mcp.tsx
```diff
@@ -0,0 +1,261 @@
...
+      clientMap.delete(clientKey(appName, appVersion));
+    };
+  }, [client, appName, appVersion]);
+
+  return (
+    <MCPClientContext.Provider value={client}>
+      {children}
+    </MCPClientContext.Provider>
+  );
+}
```
sd2k, line 224: Question for frontend folks especially: do we think this is the right abstraction? We don't _need_ to provide a provider, people can just do

```ts
const client = new Client({ appName, appVersion });
await client.connect(new GrafanaLiveTransport());
```

but the provider and hook feels quite nice to avoid too much prop drilling, avoid `useAsync`, and make use of `Suspense`.

cc @manderson-dev @obetomuniz and possibly @SandersAaronD

obetomuniz: I'd recommend sticking with the provider pattern. It slightly reduces flexibility but significantly cuts down on boilerplate, avoids prop drilling, and discourages inconsistency across the codebase. It also allows making use of Suspense/ErrorBoundary. 

Still, if I understood correctly, if a specific need for flexibility arises, I think the `GrafanaLiveTransport` export covers that case neatly. Thoughts?

sd2k: Thanks @obetomuniz! The only concern I have is that the client will be shared between different apps and versions but I can't think of many situations where that would be a real issue, so let's stick with it.

csmarchbanks (approved):
:+1: With the helpers marked as experimental I am happy to merge this and get it used in more places.

obetomuniz (approved):

PR #595, opened 2025-03-13
Allow enabling of dev sandbox via jsonData

csmarchbanks:
The dev sandbox can be helpful in deployed instances to verify if the LLM is working, and soon to see which MCP tools are available. Allow turning it on in dev or other deployed instances via jsonData rather than requiring a development build.


sd2k (approved):
lgtm!

PR #596, opened 2025-03-14
Bump plugin version to 0.14.0

sd2k:
This version includes the MCP server in the backend, behind a feature
flag.

This commit doesn't bump the frontend version; that will be done after
https://github.com/grafana/grafana-llm-app/pull/590 is merged.


sd2k:
I managed to merge the frontend PR first so this now includes a frontend version bump too.

PR #601, opened 2025-03-19
workaround: send publish messages over Websocket

sd2k:
The Grafana Live service API for publishing messages sends a message to Grafana's HTTP API rather than over the live channel, for reasons that are unclear (but presumably justified in the default case). This is fine when there is only one Grafana instance, but when there are multiple (e.g. in a HA setup), the HTTP request will be routed to a random Grafana instance, while we need it to be routed to the same instance that the client is connected to (since there is a long-lived stream over the live channel).

To work around this, we can drill down into the implementation details of the Grafana Live service and use the Centrifuge API directly to publish the message to the same stream that the client is connected to.

This works in both single-instance and HA setups, but it's not ideal, since Grafana's Live implementation is not public and could in theory change in the future. Note that the implementation has been stable for several years, so this isn't _too_ heinous. Hopefully it won't be required for long though.

TODO: replace this workaround with proper use of the public API, as in https://github.com/grafana/grafana-llm-app/pull/599, once https://github.com/grafana/grafana/pull/102325 is merged.


csmarchbanks (approved):
I hate it but would like our features to work, so let's do it.

sd2k:
I'm going to get this in now, and we can improve if/when the upstream Grafana PR is merged.

PR #602, opened 2025-03-19
feat: add tool call support to Anthropic provider

### LINKED ISSUE #600: Tool calling broken for Anthropic endpoints
We need to pass tool calls through to the Anthropic endpoints here:
https://github.com/grafana/grafana-llm-app/blob/11dae14ec87f41afe8e8478d6eaeafc1bb318a4a/packages/grafana-llm-app/pkg/plugin/anthropic_provider.go#L124-L129

and convert the response back from Anthropic to OpenAI here:
https://github.com/grafana/grafana-llm-app/blob/11dae14ec87f41afe8e8478d6eaeafc1bb318a4a/packages/grafana-llm-app/pkg/plugin/anthropic_provider.go#L91-L111

Similarly for streaming here:
https://github.com/grafana/grafana-llm-app/blob/11dae14ec87f41afe8e8478d6eaeafc1bb318a4a/packages/grafana-llm-app/pkg/plugin/anthropic_provider.go#L155-L159

https://github.com/grafana/grafana-llm-app/blob/11dae14ec87f41afe8e8478d6eaeafc1bb318a4a/packages/grafana-llm-app/pkg/plugin/anthropic_provider.go#L183-L202

Note that function calling works for OpenAI endpoints because we proxy the `ChatCompletionRequest` as is from the clients.

sd2k:
This was mostly straightforward except for converting the result
of a streaming request, which it turns out is way more complicated
than our initial implementation accounted for. Fortunately both
Anthropic and OpenAI have relatively similar ways of formatting
streaming tool calls: the first chunk/block contains the function
ID ane name, and the subsequent chunks/blocks contain partial JSON
strings containing the arguments.

Resolves #600.

### DIFF
--- a/packages/grafana-llm-app/pkg/plugin/anthropic_provider.go
+++ b/packages/grafana-llm-app/pkg/plugin/anthropic_provider.go
@@ -2,6 +2,7 @@ package plugin
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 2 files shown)


sd2k:
This is quite a lot of effort and kinda fiddly even before I get streaming working, I'm questioning whether we should just use Anthropic's OpenAI compatibility URL instead...

annanay25:
superseded by https://github.com/grafana/grafana-llm-app/pull/617

PR #603, opened 2025-03-20
fix: use public API to publish, where possible

sd2k:
https://github.com/grafana/grafana-llm-app/pull/601 added a hacky solution to publish messages over the Grafana Live Websocket connection rather than the HTTP API, making it work better in HA setups, but it accesses internal APIs so is not a long term solution.

Fortunately, https://github.com/grafana/grafana/pull/102325 has now been merged and will be available in Grafana >=11.6.0, so we can use the public API for all future versions of Grafana and only require the internal API for older versions.

This commit does that, using the public API when the `publish` function has the new signature (3 arguments) and only falling back to accessing the internal APIs if that isn't available.

### DIFF
--- a/packages/grafana-llm-frontend/src/mcp.tsx
+++ b/packages/grafana-llm-frontend/src/mcp.tsx
@@ -99,36 +99,48 @@ export class GrafanaLiveTransport implements Transport {
     // same instance that the client is connected to (since there is a
     // long-lived stream over the live channel).
     //
-    // To work around this, we can drill down into the implementation details
-    // of the Grafana Live service and use the Centrifuge API directly to
-    // publish the message to the same stream that the client is connected to.
-    //
-    // This works in both single-instance and HA setups, but it's not ideal,
-    // since Grafana's Live implementation is not public and could in theory
-    // change in the future. Note that the implementation has been stable for
-    // several years, so this isn't _too_ heinous. Hopefully it won't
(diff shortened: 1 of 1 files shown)


packages/grafana-llm-frontend/src/mcp.tsx
```diff
@@ -99,36 +99,48 @@ export class GrafanaLiveTransport implements Transport {
...
-    // TODO: replace this workaround with proper use of the public API,
-    // as in https://github.com/grafana/grafana-llm-app/pull/599,
-    // once https://github.com/grafana/grafana/pull/102325 is merged.
+    // We can use the `useSocket` argument when trying to publish to the
+    // live channel to force the use of the Websocket instead of the HTTP API.
+    // This will work in both single-instance and HA setups. However, it's only
+    // available in Grafana 11.6.0 and later. We can check for this by checking
+    // if the `publish` method has a third argument, which is the `options`
+    // argument.
+    const hasPublishOptions = this._grafanaLiveSrv.publish?.length === 3;
```
gerboland, line 108: That's a very particular check! Why 3? Could this change to >3 in future?

sd2k: Good call, for posterity this was fixed in https://github.com/grafana/grafana-llm-app/pull/604.

PR #604, opened 2025-03-21
fix: check for >= 3 args in publish, not === 3

sd2k:
It's unlikely but the signature may change in future and add a further
argument, and we should handle that case.


gerboland (approved):
thanks! :)

PR #612, opened 2025-03-27
Bump a full minor version to do a release so we can publish a version out of public preview

SandersAaronD:
Necessary to finalize pr #610 on the web site and in the release cycle


annanay25 (approved):
lgtm

PR #622, opened 2025-04-02
Update release instructions for NPM package

annanay25:
~Noticed we are not using the old notation anymore, updating instructions to just use `XX.XX.XX`~

`llmclient` release instructions were correct, adding an additional note for NPM release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,12 @@ It is recommended to run these using npm commands from root so that the entire p
 - Bump version in `packages/grafana-llm-app/package.json` (e.g., 0.2.0 to 0.2.1)
 - Add notes to changelog describing changes since last release
 - Merge PR for a branch containing those changes into main
-- Trigger release from "actions" tab on Github.
+- Trigger "Plugin Release" from the Actions tab on Github.
+
+### NPM Release
+This is for the npm package `@grafana/llm`.
+- Bump version in `packages/grafana-llm-frontend/package.json` (e.g., 0.2.0 to 0.2.1)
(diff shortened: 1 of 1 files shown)


sd2k (commented):
I actually haven't been doing this, I've just been manually triggering the [NPM release](https://github.com/grafana/grafana-llm-app/actions/workflows/npm-release.yml) GitHub action from `main` after updating the version in package.json. The instructions need updating either way!

csmarchbanks:
Yeah, I think creating a tag might actually break the action and not publish the release so we should update the docs to reflect that.

README.md
```diff
@@ -112,4 +112,4 @@ It is recommended to run these using npm commands from root so that the entire p
 - Trigger release from "actions" tab on Github.
 
 ### llmclient Release
-- Push a new tag to the repo (e.g., `git tag -a llmclient/v0.X.X -m "llmclient v0.X.X release"`)
```
csmarchbanks, line 115: I think the llmclient prefix is required for go mod? Or is that not the case anymore?

csmarchbanks:
Ohh wow, yeah this is woefully out of date: https://pkg.go.dev/github.com/grafana/grafana-llm-app/llmclient. Do we want the go client version to be aligned with the rest of the repo?

sd2k:
Hmm, most of the version bumps will be no-ops for that package which might be a bit strange for users who have e.g. dependabot running. OTOH it's annoying to have to keep track of multiple version numbers...

csmarchbanks (approved):

Issue #625, opened 2025-04-10
Handle inconsistency between Anthropic and OpenAI API behaviours


edwardcqian:
OpenAI API allows a request to be sent without a user message (i.e. only system prompts).

However, anthropic does not allow that.

This results in potential issues in features if providers are swapped.

For example:

Flamegraph explainer is broken is ops due to ops using anthropic (since it only send system prompts)

<img width="796" alt="Image" src="https://github.com/user-attachments/assets/2d397038-4fbe-4516-a105-a2335f88900b" />

A simple way to handle this might just be adding some a meaningless user message if one doesn't already exist. (double checked in the anthropic playground that the request works and a valid answer is still provided.)

PR #629, opened 2025-04-14
Anthropic provider: add logic to force user message

edwardcqian:
fixes: #625 

Test by changing DevSandBox

```
{ role: 'user', content: message },
```

To system role

### DIFF
--- a/packages/grafana-llm-app/pkg/plugin/anthropic_provider.go
+++ b/packages/grafana-llm-app/pkg/plugin/anthropic_provider.go
@@ -57,6 +57,10 @@ func (p *anthropicProvider) Models(ctx context.Context) (ModelResponse, error) {
 }
 
 func (p *anthropicProvider) forceUserMessage(req *openai.ChatCompletionRequest) {
+	if len(req.Messages) == 0 {
+		return
+	}
+
 	hasUserMessage := false
 	for _, message := range req.Messages {
 		if message.Role == "user" {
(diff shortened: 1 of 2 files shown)


packages/grafana-llm-app/pkg/plugin/anthropic_provider.go
```diff
@@ -56,6 +56,20 @@ func (p *anthropicProvider) Models(ctx context.Context) (ModelResponse, error) {
...
+	hasUserMessage := false
+	for _, message := range req.Messages {
+		if message.Role == "user" {
+			hasUserMessage = true
+			break
+		}
+	}
+
+	if !hasUserMessage {
+		req.Messages[len(req.Messages)-1].Role = "user"
```
csmarchbanks, line 69: Could we short circuit (and possibly return error?) if there are no messages?

edwardcqian: Going to just return if there are no messages.

Throwing error on empty messages should probably be handled else where in the plugin.

We currently rely on features themselves to handle errors coming from the APIs and when they don't handle it, the frontend just hangs.

csmarchbanks: Just returning sounds good to me, wouldn't change current behavior and avoids the panic/DOS issue.

csmarchbanks (approved):

PR #632, opened 2025-04-17
Update default OpenAI models to GPT 4.1

csmarchbanks:
We are seeing significant improvements when using GPT 4.1 compared to 4o in our internal benchmarks.


sd2k (approved):
LGTM

csmarchbanks:
Hmm, yeah we should think about it in 611, there are other things to consider as well such as if it is the default should we upgrade them? Our managed one is probably the only safe one to do that for.

PR #643, opened 2025-04-24
feat!: add enabled flag to MCPProvider context value

sd2k:
Rather than just returning the client instance, the context value now
includes a boolean flag indicating whether MCP is enabled for the current
Grafana instance. This is useful for cases where the LLM plugin is not
installed, or the user has disabled MCP; previously the provider would
throw an error in these cases resulting in an infinite loop.

This is a breaking change since the context value is now an object with
an `enabled` flag.


packages/grafana-llm-frontend/src/mcp.tsx
```diff
@@ -206,6 +240,10 @@ function createClientResource(appName: string, appVersion: string): ClientResour
   return {
     read() {
       if (status === 'pending') {
+        if (!result?.enabled) {
+          console.warn('MCP is not enabled');
+          return result || { client: null, enabled: false };
+        }
         throw promise;
```
obetomuniz, line 247: @sd2k tested locally the changes, and found out that our problem is here. I added this logic just to markup, but the fact is that when the client returns 404 (e.g. MCP not enabled), it enters in an infinite state of `pending`. The problem with this `!result?.enabled` check is that this is async, so not reliable as its initial state will be always `null | false`.

obetomuniz, line 247: @sd2k tested locally the changes, and found out that our problem is here. I added this logic just to markup, but the fact is that when the client returns 404 (e.g. MCP not enabled), it enters in an infinite state of pending. The problem with this !result?.enabled check is that this is async, so not reliable as its initial state will be always null | false, which give us a false-positive when MCP is enabled.

obetomuniz:
Just pushed afdfd07, which turned `createClientResource` function into an async function, which allows us to update the check whether MCP is enabled or not after the request is done and before it starts the infinite loop when MCP is not enabled in the environment. Tested locally pointing to different environments, and that's the results:

asserts.grafana-ops.net (MCP disabled)

https://github.com/user-attachments/assets/16fead00-69a9-457b-a596-9f314667cff0

The answer is textual and based on the data injected via prompt to the LLM.

assertschief.grafana-dev.net (MCP enabled)

https://github.com/user-attachments/assets/cadbf1cf-1f00-4ed2-9f71-a75adabf5dbe

The answered is structured in a custom UI and based on MCP tools enabled in the MCP server.

--- 

I think that's a good result, since we now can use the `mcp.MCPClientProvider` provider without degradation when MCP is not enabled.

packages/grafana-llm-frontend/src/mcp.tsx
```diff
@@ -259,25 +294,25 @@ export function MCPClientProvider({
...
+        result.client.close();
       }
       clientMap.delete(clientKey(appName, appVersion));
     };
-  }, [client, appName, appVersion]);
+  }, [result, appName, appVersion]);
 
   return (
-    <MCPClientContext.Provider value={client}>
+    <MCPClientContext.Provider value={result ?? null}>
```
sd2k, line 315: if we return `null` here would this not trigger the `if (client === null)` check on line 330?

sd2k:
I'm still a bit confused about the behaviour before your most recent commit :thinking: My intended flow was that if the MCP client was still loading (i.e. checking whether enabled _or_ connecting - both async operations) then promise would be thrown which would trigger the parent `<Suspense />` to be run. It feels like that won't be the case with the most recent commit though: instead we're setting the Provider's `value` to `null` during the loading state?

sd2k:
I wonder if https://github.com/grafana/grafana-llm-app/pull/648 works for your use case @obetomuniz? The only real difference is https://github.com/grafana/grafana-llm-app/pull/648/commits/65a0d235be4665638fb05b1d0d6b84452c72de7d.

obetomuniz:
@sd2k tested locally this change, but it has the issues I noticed before my changes. Below you can see the code I am using to setup it with Suspense and demos:

```tsx
<Suspense fallback={<FallbackLoadingPlaceholder />}>
  <ErrorBoundary>
    {({ error }) => {
      if (error) {
        return <div>Something went wrong: {error.message}</div>;
      }
      return (
        <mcp.MCPClientProvider appName="grafana-assertions" appVersion="1.0.0">
          <IntlProvider locale="en">
            <Provider store={store}>
              <OutsideCallConsumer config={callConfig}>
                <PersistGate loading={null} persistor={persistor}>
                  <Suspense fallback={<FallbackLoadingPlaceholder />}>
                    <AppRoutes />
                  </Suspense>
                </PersistGate>
              </OutsideCallConsumer>
            </Provider>
          </IntlProvider>
        </mcp.MCPClientProvider>
      );
    }}
  </ErrorBoundary>
</Suspense>
```

When pointing to `assertschief.grafana-dev.net` (MCP enabled) the MCP tools doesn't work since the `isEnabled` is async and starts with `null`, and early return since it falls into the `status === 'pending'` condition without updating the state of `isEnabled` to `true`.

https://github.com/user-attachments/assets/0ddf43df-9e5e-4587-a7e5-07872a9ed8be

When pointing to `asserts.grafana-ops.net` (MCP disabled). The LLM response is not accurate, it is just an hallucination from that env:

https://github.com/user-attachments/assets/a4670a84-172b-446a-9384-aa967cd1998c

obetomuniz:
Closing. https://github.com/grafana/grafana-llm-app/pull/648 addressed the problem.

Issue #593, opened 2025-03-13
MCP: use on-behalf-of auth rather than service account when possible


sd2k:
Maybe break this into an issue rather than code comments and just link to the issue?

_Originally posted by @csmarchbanks in https://github.com/grafana/grafana-llm-app/pull/574#discussion_r1993695624_

PR #644, opened 2025-04-25
Enable LLM app to create access tokens on-behalf-of user

annanay25:
depends on https://github.com/grafana/mcp-grafana/pull/107 and releasing a new version of `mcp-grafana`

fixes https://github.com/grafana/grafana-llm-app/issues/593


packages/grafana-llm-app/pkg/mcp/live_server.go
```diff
@@ -94,14 +124,42 @@ type liveSession struct {
...
 
 // HandleStream handles a new Grafana Live session.
 func (s *GrafanaLiveServer) HandleStream(ctx context.Context, req *backend.RunStreamRequest, sender *backend.StreamSender) error {
+	var tokenExchangeResponse *authn.TokenExchangeResponse
+	// Only do the token exchange if we are in Grafana Cloud.
+	if s.EnableGrafanaManagedLLM {
+		var tokenExchangeClient *authn.TokenExchangeClient
+		var err error
+		// We need to use the token exchange API to create a token for the Grafana Live session.
+		tokenExchangeClient, err = authn.NewTokenExchangeClient(authn.TokenExchangeConfig{
```
cinaglia, line 140: I could be missing something obvious but generally speaking, you'll want to instantiate the client once only somewhere as high up in the chain as possible so that you can benefit from the caching strategy.

packages/grafana-llm-app/pkg/mcp/live_server.go
```diff
@@ -94,14 +124,42 @@ type liveSession struct {
...
+			TokenExchangeURL: "http://api-lb.auth.svc.cluster.local./v1/sign-access-token", // TODO: make this configurable.
+		})
+		if err != nil {
+			return err
+		}
+
+		tokenExchangeResponse, err = tokenExchangeClient.Exchange(ctx, authn.TokenExchangeRequest{
+			Namespace: fmt.Sprintf("stack-%s", s.Tenant),
+			Audiences: []string{"grafana"},
+			ExpiresIn: &[]int{600}[0], // this is 10 minutes
```
cinaglia, line 151: nit: You don't need to provide this if you're OK with the 10 minute expiration which is set by default.

annanay25:
queued behind https://github.com/grafana/grafana/pull/104833

packages/grafana-llm-app/pkg/mcp/live_server.go
```diff
@@ -145,9 +220,30 @@ func (s *GrafanaLiveServer) HandleMessage(ctx context.Context, req *backend.Publ
...
+
+	// If we are not using Grafana Cloud, use the API key. In this case, the
+	// access token and grafana id token will be empty.
+	if accessToken == "" || grafanaIdToken == "" {
+		apiKey, _ := cfg.PluginAppClientSecret()
+		return mcpgrafana.WithGrafanaAPIKey(mcpgrafana.WithGrafanaURL(ctx, url), apiKey)
+	}
+
+	// If we have an access token and grafana id token, use on-behalf-of auth.
+	return mcpgrafana.MustWithOnBehalfOfAuth(mcpgrafana.WithGrafanaURL(ctx, url), accessToken, grafanaIdToken)
```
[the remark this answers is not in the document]
annanay25: good call, fixed 👍

packages/grafana-llm-app/pkg/mcp/live_server.go
```diff
@@ -67,16 +82,45 @@ func WithGrafanaLiveContextFunc(contextFunc GrafanaLiveContextFunc) GrafanaLiveO
...
 	for _, opt := range opts {
 		opt(s)
 	}
-	return s
+
+	if s.enableGrafanaManagedLLM {
+		var err error
+		s.tokenExchangeClient, err = authn.NewTokenExchangeClient(authn.TokenExchangeConfig{
+			Token:            s.llmAppAccessPolicyToken,
+			TokenExchangeURL: "http://api-lb.auth.svc.cluster.local./v1/sign-access-token", // TODO: make this configurable.
```
sd2k, line 117: Can this be obtained from the environment or something similar?

packages/grafana-llm-app/pkg/mcp/live_server.go
```diff
@@ -46,6 +52,15 @@ type GrafanaLiveContextFunc func(ctx context.Context, pCtx *backend.PluginContex
 type GrafanaLiveServer struct {
 	// server is the MCP server that will handle the MCP messages.
 	server *server.MCPServer
+	// tenant is the stack ID (Hosted Grafana ID) of the instance this plugin
+	// is running on.
+	tenant string
+	// llmAppAccessPolicyToken is the token created from the LLM app's access policy.
+	llmAppAccessPolicyToken string
+	// enableGrafanaManagedLLM is a flag to indicate if this is running in Grafana Cloud.
+	enableGrafanaManagedLLM bool
```
sd2k, line 61: I'm a bit confused about the intention here. I guess we need to gate some things on whether we're in Grafana Cloud, and we're using the Grafana Managed LLM setting as a proxy for that? IMO we should rename this field to `isGrafanaCloud` in this package and add a comment at the creation site stating that we're using the Grafana Managed LLM setting as a proxy for that (and/or search for a better indicator of it being Grafana cloud 🙂)

packages/grafana-llm-app/pkg/mcp/live_server.go
```diff
@@ -22,6 +25,9 @@ const (
 	subscribeSuffix = "/subscribe"
 	// publishSuffix is the suffix for the publish channel endpoint.
 	publishSuffix = "/publish"
+
+	// AccessTokenHeader is the HTTP header key for the access token.
+	AccessTokenHeader = "X-Access-Token"
```
sd2k, line 30: nit: doesn't need to be exported

packages/grafana-llm-app/pkg/mcp/live_server.go
sd2k: just did some searching and I don't think it is 🤷 so no biggie!

sd2k:
Looks like we need a new release of mcp-grafana then we can get tests passing here too 👍

csmarchbanks:
I released mcp-grafana just a bit ago

packages/grafana-llm-app/pkg/mcp/live_server.go
csmarchbanks: Yeah, I agree that we should not propagate the same name used in config here. Ideally we would rename the config variable to be honest, but that involves reconciling a lot of Grafana Cloud stacks.

annanay25: I've added a clarifying comment above the variable now. I think we should keep the same name within `live_server.go` so its clear at the time of refactor?

sd2k: Ah I was thinking we would rename it only in this file, but honestly it's not a huge deal, we can do it later. Approving to get you unblocked!

sd2k (approved):

PR #648, opened 2025-04-28
feat!: add enabled flag to MCPProvider context value (take 2)

### LINKED ISSUE #637: `mcp.MCPClientProvider` hangs when MCP is not enabled
While using the `mcp.MCPClientProvider` we don't see any errors/response if MCP is not enabled. Ideally we would have another helper function to expose the enabled setting.

cc @kozhuhds @obetomuniz

sd2k:
Mostly the same as #643 up to the most recent commits, but branched off with a couple of bugfixes (I hope).

Fixes #637

### DIFF
--- a/packages/grafana-llm-app/src/components/AppConfig/DevSandbox.tsx
+++ b/packages/grafana-llm-app/src/components/AppConfig/DevSandbox.tsx
@@ -259,7 +259,7 @@ const BasicChatTest = () => {
       return { enabled, tools: [] };
     }
 
-    const { tools } = await client.listTools();
+    const { tools } = await client?.listTools() ?? { tools: [] };
     if (message === '') {
       return { enabled, tools };
     }
--- a/packages/grafana-llm-app/src/components/AppConfig/DevSandbox.tsx
+++ b/packages/grafana-llm-app/src/components/AppConfig/DevSandbox.tsx
@@ -238,7 +238,7 @@ function ToolCalls({ toolCalls }: { toolCalls: Map<string, RenderedToolCall> }) 
 }
 
 const BasicChatTest = () => {
-  const client = mcp.useMCPClient();
+  const { client } = mcp.useMCPClient();
   // The current input value.
   const [input, setInput] = useState('');
   // The final message to send to the LLM, updated when the button is clicked.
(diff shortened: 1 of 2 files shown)


obetomuniz:
@sd2k this branch fixes the infinite loop when using `mcp.MCPClientProvider` in an instance without MCP enabled and make this provider work properly with Suspense without need to convert `createClientResource` to async. Below you can see the code I am using:

```tsx
<Suspense fallback={<FallbackLoadingPlaceholder />}>
  <ErrorBoundary>
    {({ error }) => {
      if (error) {
        return <div>Something went wrong: {error.message}</div>;
      }
      return (
        <mcp.MCPClientProvider appName="grafana-assertions" appVersion="1.0.0">
          <IntlProvider locale="en">
            <Provider store={store}>
              <OutsideCallConsumer config={callConfig}>
                <PersistGate loading={null} persistor={persistor}>
                  <Suspense fallback={<FallbackLoadingPlaceholder />}>
                    <AppRoutes />
                  </Suspense>
                </PersistGate>
              </OutsideCallConsumer>
            </Provider>
          </IntlProvider>
        </mcp.MCPClientProvider>
      );
    }}
  </ErrorBoundary>
</Suspense>
```

Pointing to `assertschief.grafana-dev.net` (MCP enabled), it invokes and renders correctly the MCP tool call.

https://github.com/user-attachments/assets/3dd2c455-7762-44b8-8423-f40a112ea3b9

Pointing to `asserts.grafana-ops.net` (MCP disabled), it gracefully renders the UI without hangs, even when the user attempts to use some MCP tools. 

https://github.com/user-attachments/assets/71e4a0fe-65d1-46e6-939d-221134fbc6af

I think we can close #643 and open this PR to be merged.  

I wonder that would be nice to expose `mcp.enabled()` function, so we can use it programmatically as necessary, like tell the user that the environment doesn't have MCP enabled and things like that. Or even use specific MCP tools interactions only when it is fully setup. Thoughts?

obetomuniz (approved):

obetomuniz:
Hey @sd2k, are there still issues with GitHub Actions that affect the merge and release of this PR?

PR #670, opened 2025-05-16
Remove MCP feature flag and enable it by default

### LINKED ISSUE #666: Enable Grafana MCP server by default
Now that we have on-behalf-of-auth implemented as part of #644, we should drop the requirement to have `{mcp: enabled}` set in the plugin JSON data.

annanay25:
Fixes #666


annanay25:
@sd2k is it safe to drop config settings in one go? wondering if JSON parsing is strict and will crash if we still have `{mcp: {enabled: true}}` set on an instance.

sd2k:
Extra JSON should be dropped without an issue unless you use a decoder with [`DisallowUnknownFields`](https://pkg.go.dev/encoding/json#Decoder.DisallowUnknownFields) set, so we should be fine. 

I'm not sure if we should keep this enabled option around in case users want to disable it though... Is there any good reason for that?

annanay25:
> I'm not sure if we should keep this enabled option around in case users want to disable it though... Is there any good reason for that?

Hmm, I think there might be a case for that. Pricing concerns / query volumes / etc?

annanay25:
switched this to a disabled flag instead.

sd2k (approved):

PR #682, opened 2025-05-28
chore: use vault secrets instead of GitHub secrets

sd2k:
GitHub secrets will be removed soon; we should fetch secrets from Vault instead.

I've added the relevant repo secrets to Vault but had to regenerate a cloud access policy token for this plugin. Hard to say if it will all work smoothly until we do a release, unfortunately.

### DIFF
--- a/.github/workflows/plugin-release.yml
+++ b/.github/workflows/plugin-release.yml
@@ -182,12 +182,6 @@ jobs:
             app_id=grafana-machine-learning-github-app:app-id
             app_installation_id=grafana-machine-learning-github-app:app-installation-id
             private_key=grafana-machine-learning-github-app:private-key
-          export_env: false
-
-      - id: get-access-policy-secret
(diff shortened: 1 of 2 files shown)


.github/workflows/plugin-release.yml
```diff
@@ -180,6 +185,13 @@ jobs:
             private_key=grafana-machine-learning-github-app:private-key
           export_env: false
 
+      - id: get-access-policy-secret
+        uses: grafana/shared-workflows/actions/get-vault-secrets@5d7e361bc7e0a183cde8afe9899fb7b596d2659b
+        with:
+          repo_secrets: |
+            GRAFANA_CLOUD_ACCESS_POLICY_TOKEN=grafana-cloud-access-policy:token
```
csmarchbanks, line 192: Can't this just be added to the above step? Any reason to keep them separate?

csmarchbanks (approved):

PR #688, opened 2025-06-04
feat: refresh Grafana access token before expiry in Cloud

sd2k:
Prior to this commit, the Grafana Live session handler would not refresh
the access token before it expired. This causes issues when using long
running streams in Grafana, as the token would expire halfway through
the stream and the client would experience strange errors.

This commit adds a timer to the Grafana Live session handler to refresh
the access token one minute before it expires. On refresh, the timer
will reset to fire one minute before the token is due to expire again.

### DIFF
--- a/packages/grafana-llm-app/pkg/mcp/live_server.go
+++ b/packages/grafana-llm-app/pkg/mcp/live_server.go
@@ -150,13 +150,14 @@ type liveSession struct {
 }
 
 // tokenTimeoutSeconds is the expiration time for the Grafana Live session token.
-var tokenTimeoutSeconds = int((time.Minute * 30).Seconds())
+var tokenTimeout = time.Minute * 30
 
 // tokenRefreshInterval is the time between token refreshes.
-var tokenRefreshInterval = time.Duration(tokenTimeoutSeconds) - time.Minute
+var tokenRefreshInterval = tokenTimeout - time.Minute
 
 // exchangeToken uses the token exchange API to create an access token for the Grafana Live session.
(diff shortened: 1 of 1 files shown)


packages/grafana-llm-app/pkg/mcp/live_server.go
```diff
@@ -148,24 +149,44 @@ type liveSession struct {
 	grafanaIdToken string
 }
 
+// tokenTimeoutSeconds is the expiration time for the Grafana Live session token.
+var tokenTimeoutSeconds = int((time.Minute * 30).Seconds())
+
+// tokenRefreshInterval is the time between token refreshes.
+var tokenRefreshInterval = time.Duration(tokenTimeoutSeconds) - time.Minute
```
csmarchbanks, line 156: Doesn't this need to be something like the following?
```suggestion
var tokenRefreshInterval = time.Duration(tokenTimeoutSeconds) * time.Second - time.Minute
```

packages/grafana-llm-app/pkg/mcp/live_server.go
```diff
@@ -148,24 +149,44 @@ type liveSession struct {
 	grafanaIdToken string
 }
 
+// tokenTimeoutSeconds is the expiration time for the Grafana Live session token.
+var tokenTimeoutSeconds = int((time.Minute * 30).Seconds())
```
csmarchbanks, line 153: I think it would be better if this was a time.Duration and we do the slightly annoying make a new variable when we do the token exchange. That way we don't have to deal with as many unit conversions.

PR #689, opened 2025-06-04
feat: export 'enabled' function from mcp module of npm package

sd2k:
This will be useful for anyone using the MCP server outside of a
React component context.

### DIFF
--- a/packages/grafana-llm-frontend/src/mcp.tsx
+++ b/packages/grafana-llm-frontend/src/mcp.tsx
@@ -181,7 +181,12 @@ type ClientResource = {
   read: () => ClientResult;
 };
 
-async function isEnabled(): Promise<boolean> {
+/**
+ * Check if the Grafana LLM app is installed and the MCP server is enabled for the current Grafana instance.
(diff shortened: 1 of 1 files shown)


packages/grafana-llm-frontend/src/mcp.tsx
```diff
@@ -181,7 +181,12 @@ type ClientResource = {
...
-async function isEnabled(): Promise<boolean> {
+/**
+ * Check if the Grafana LLM app is installed and the MCP server is enabled for the current Grafana instance.
+ *
+ * @returns Whether MCP is enabled for the current Grafana instance.
+ */
+export async function enabled(): Promise<boolean> {
   try {
     const settings = await getBackendSrv().get(`${LLM_PLUGIN_ROUTE}/settings`, undefined, undefined, {
       showSuccessAlert: false,
```
annanay25, line 192: we changed the setting to a `disabled` flag - do we need to change that here?

annanay25 (approved):
lgtm!

PR #691, opened 2025-06-04
PoC: add MCP StreamableHTTP server to /mcp endpoint of LLM app

sd2k:
This is a PoC demonstrating how we could use the StreamableHTTP
transport with our Grafana plugin's resource endpoints to serve
the Grafana MCP server at /mcp.

You can call the endpoints using something like:

```sh
curl localhost:3000/api/plugins/grafana-llm-app/resources/mcp \
  -H 'content-type: application/json' \
  -d '{"jsonrpc": "2.0", "method": "tools/list"}'
```

(or using a proper MCP client).

This is still a PoC and doesn't handle auth, but it wouldn't be too hard
to do something similar to the Grafana Live server to get that working
using OBO auth.


sd2k:
Marking as draft, I'd like to rethink the `/mcp` endpoints to potentially accommodate multiple MCP servers.

packages/grafana-llm-app/pkg/plugin/resources.go
```diff
@@ -600,4 +600,8 @@ func (a *App) registerRoutes(mux *http.ServeMux) {
 	mux.HandleFunc("/vector/search", a.handleVectorSearch)
 	mux.HandleFunc("/grafana-llm-state", a.handleLLMState)
 	mux.HandleFunc("/save-plugin-settings", a.handleSavePluginSettings)
+
+	if a.mcpServer != nil {
+		mux.HandleFunc("/mcp", a.mcpServer.HTTPServer.ServeHTTP)
```
sd2k, line 605: E.g. perhaps this should be on `/mcp/grafana`

annanay25 (commented):
lgtm on the PR, just trying to figure out why the frontend test is failing.

sd2k:
Thanks @annanay25! Those jest ignores worked a treat, I tried and failed to get that working, you're a champ.

This is now ready for review, if anyone is able to take a look.

sd2k:
Hmm gonna rebase on main to check e2e tests still pass.

PR #696, opened 2025-06-04
refactor: add MCP struct and simplify access token usage

sd2k:
This commit adds a new MCP struct that encapsulates the MCP server and Grafana Live server, as well as holding all MCP-related settings. Finally, it adds a new access token client that encapsulates the access token exchange process, including skipping it if we are not in Grafana Cloud.

Rather than manually refreshing access tokens, this relies on the fact that the authlib library's TokenExchangeClient will automatically cache and refresh the access token if it is expired.

Note: the restructuring of the MCP side is partly motivated by wanting to also use the access token client in the HTTP server in #691.

It might look like a net code addition but I think a lot of that is comments - it should be simpler to understand how!

### DIFF
--- a/packages/grafana-llm-app/pkg/mcp/accesstoken.go
+++ b/packages/grafana-llm-app/pkg/mcp/accesstoken.go
@@ -1,0 +1,65 @@
+package mcp
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/grafana/authlib/authn"
+)
+
+// tokenTimeout is the expiration time for the Grafana Live session token.
+// Note: We tried setting this to 30 minutes, but it caused the following error:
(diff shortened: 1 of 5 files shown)


sd2k:
Might be worth waiting until #701 is in before doing this so we're sure everything still works correctly.

SandersAaronD (approved):
This looks good to me, I would be good with waiting for the playwright PR to land and checking to make sure nothing breaks due to changes there before merging

annanay25 (approved):
lgtm. i think its ok to merge this one without Playwright tests!

sd2k:
Cool! FWIW I ran the e2e tests locally after rebasing on this branch and got matching results so I think we're good too.

PR #701, opened 2025-06-05
tests: add end-to-end testing using Playwright

sd2k:
This commit adds an end-to-end testing infrastructure and UI improvements
to ensure the MCP functionality works as expected in a real Grafana
plugin.

To do so, I had to add a page to the plugin which actually uses the MCP
server, so there's been a bit of a rewrite of the main plugin page, but
before it just showed the models list which wasn't very useful.

I followed the [plugin tools docs](https://grafana.com/developers/plugin-tools/e2e-test-a-plugin/migrate-from-grafana-e2e)
for migrating from the old e2e tests (not that we had any anyway), so
many of the generated files come from that, but there are also some
rather verbose files generated by Claude which we might want to remove
before merge.


sd2k:
Hm screenshots show no MCP tools, will have a look at that tomorrow

obetomuniz (approved):
LGTM :shipit: 

Just added a tiny guide to help MacOS users if they figure out port conflict issues with AirPlay Receiver.

Screenshot 2025-06-10 at 11 19 02

PR #707, opened 2025-06-06
ci: use shared GitHub actions provided by plugins platform

sd2k:
This will get us automatic builds and pushes on every PR and commit to main, more comprehensive testing and linting, the ability to use the plugin catalog, and Argo Workflows for deployment.


gitdoluquita (approved):
LGTM!
One weird thing that doens't look like it's part of this PR is the title that appears on github for these jobs:
image

Let me know if there is something I could double-check for you, that would help reviewing this.

sd2k:
We still have the old CI pipelines so it's purely additive - I'll followup with fixes if things don't work in dev :slightly_smiling_face:

PR #708, opened 2025-06-07
Fix golangci-lint errors

sd2k:


packages/grafana-llm-app/pkg/plugin/resources.go
```diff
@@ -109,7 +109,7 @@ func getLLMOptInState(ctx context.Context, settings *Settings) (llmGatewayRespon
 	if err != nil {
 		return llmGatewayResponse{}, fmt.Errorf("failed to send request to llm-gateway %w", err)
 	}
-	defer resp.Body.Close()
+	defer resp.Body.Close() //nolint:errcheck
```
sd2k, line 112: These are kind of annoying, I guess I would like to know if this goes wrong but there's not an awful lot we can do about it?

PR #713, opened 2025-06-10
Svennergr/add assitant register

svennergr:


sd2k:
@svennergr this isn't needed any more right, we have a separate NPM package?

svennergr:
Yea, sorry.

PR #717, opened 2025-06-11
fix(mcp): fix `isEnabled` check and improve back-compat

sd2k:
This PR makes the new MCP features a bit more robust for older backend versions by:

- improving the `enabled()` check to handle null cases, and adding test cases
- including an error boundary when rendering tools in the homepage
- making the dev experience and e2e tests a bit more realistic by removing the `enabled: true` flag from docker-compose provisioning (which is the default in cloud)
- correctly caching connection errors rather than entering a continuous retry loop, if the MCP endpoints aren't available (they return a 404)


annanay25 (approved):
lgtm 👍

PR #721, opened 2025-06-12
ci: add GitHub action to bump version and generate changelog

sd2k:
Note: a GitHub release will be created by the CD pipeline once we
publish to prod, not by this new action.

This was mostly copied from https://github.com/grafana/plugin-ci-workflows/blob/main/actions/plugins/version-bump-changelog/action.yml but adapted for our curious setup of having multiple packages which both need to be updated.


gerboland (approved):
looks ok, let's see how it runs

PR #742, opened 2025-06-30
fix: correctly handle model conversion for Anthropic if no mapping is provided

### LINKED ISSUE #645: Bug: Anthropic Model check succeeds with incorrect models selected
Image

While setting up a new client in the LLM plugin config page, I forgot to change the model mappings from my previous setup.

This meant that the gemini models were selected for use with the Anthropic API, which is not possible. 

What did I expect? Something like: 
```
LLM provider health check succeeded!
Models
base: Unavailable
large: Unavailable
```

What actually happened: 
```
LLM provider health check succeeded!
Models
base: OK
large: OK
```

Not a blocker, but could be confusing for a user.

### LINKED ISSUE #741: Bug: By default the anthropic provider attempts to load openai models
If you only specify the anthropic provider with no explicit models configured it will attempt to use openai models and fail with:

```
WARN [06-30|09:43:57] Unable to get service account token      logger=plugin.grafana-llm-app err="PluginAppClientSecret not set in config"
ERROR[06-30|09:43:58] error creating anthropic chat completion logger=plugin.grafana-llm-app err="error, status code: 404, status: 404 Not Found, message: model: gpt-4.1-mini"
ERROR[06-30|09:43:58] error creating anthropic chat completion logger=plugin.grafana-llm-app err="error, status code: 404, status: 404 Not Found, message: model: gpt-4.1"
```

Workaround is to explicitly set anthropic models:

Image

I believe this PR caused the issue: https://github.com/grafana/grafana-llm-app/pull/734

sd2k:
Prior to this commit we had a constant mapping of 'base' / 'large' to
OpenAI model names. This was incorrect if the user chooses Anthropic
as the provider but doesn't override the model settings.

Instead, we now parameterize the mapping based on the provider.

We also handle partial mapping overrides much better; previously we
would fall back to the configured default model if the user didn't provide
a mapping for a given model, but now we just use the default. For
example, if the user configured a mapping for 'base' but not 'large'
and made a request with 'large', we would have fallen back to the
'base' model (since it was marked as 'default'). Now we use the
fallback default model instead.

Finally, we update the UI to make it more obvious when mappings have
been set (using placeholders), and switch based on the provider, too.

Fixes #741.
Fixes #645.

### DIFF
--- a/packages/grafana-llm-app/pkg/plugin/resources_test.go
+++ b/packages/grafana-llm-app/pkg/plugin/resources_test.go
@@ -4,11 +4,14 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
+	"io"
 	"net/http"
 	"net/http/httptest"
 	"testing"
 
 	"github.com/grafana/grafana-plugin-sdk-go/backend"
+	"github.com/sashabaranov/go-openai"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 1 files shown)


sd2k:
Good call, I added some more detailed tests using a mock server to ensure things get mapped correctly depending on the mapping settings & model input.

PR #750, opened 2025-07-17
feat: Tidy up and improve dev sandbox

SandersAaronD:
This is sort of intermediate but seems worth raising: because the dev sandbox was quite cluttered it was difficult to actually see what MCP tools you had, this makes it a more standard chat window with a separate modal within it for examining MCP tools.

I'd rather see this merged before continuing but if it hangs for a while I'll just add more logic to it where hopefully we can e2e test through the dev sandbox


sd2k (approved):
A modal on a modal is a bit weird 😅 I wonder if we should use [Tabs](https://developers.grafana.com/ui/canary/index.html?path=/docs/layout-tabs--docs) and have the tool inspector in a tab instead? I would also love it if the tool calls appeared inline in the chat rather than in a further modal but it's not worth too much time.

Approving anyway because this is definitely an improvement.

Issue #760, opened 2025-07-25
Grafana LLM App Admin Page UX Issues and Bugs


manoja1:
Please watch the recording for the UX issues and Bugs. In summary

- It's not evident what's in use - Grafana's own vs Bring your own
- MOdel name section should belong to Bring your own
- Save & Test results in wrong configuration being saved when the test fails later
- Switching between models doesn't prompt me to update the model name

https://github.com/user-attachments/assets/a660552b-6be1-4564-a1f9-44804e8f37b8

cc: @Maurice-L-R @devinrcheevers @brittenhouse @vaugun @okianus

sd2k:
@iamrajiv we should use a separate issue for Gemini stuff! There is [this existing one](https://github.com/grafana/grafana-llm-app/issues/628) which might suffice.

I'll take a look at your PR soon, thanks!

iamrajiv:
sure thanks @sd2k

manoja1:
I also sometimes get this error when I do Save & Test .. and takes 5/10 hard resets before it goes away

<img width="731" height="222" alt="Image" src="https://github.com/user-attachments/assets/f8977ebf-8bba-436f-bbd1-1e49b69368ba" />

sd2k:
@vaugun one comment based on this issue and the draft PR by @iamrajiv - it makes sense to group together the 'Model provider' and 'Models' sections more tightly, because you're likely to need to change the latter whenever you change the former. The second screenshot in #762 shows an example.

Otherwise I think it looks good 👍 thanks!

vaugun:
@sd2k Roger that! I'll update accordingly 🫡

jtheory:
Once we've made a new release, this is worth another check to see if we should pause where we are now, or if some of @vaugun's suggestions will be a step better; if so we can leave this issues open with clearer next steps.

PR #762, opened 2025-07-26
feat(ui): Admin page UX fixes

iamrajiv:
Fixes: #760

before:

- no clear provider indication
- models section at bottom
- no switching warnings
  etc

after:

Screenshot 2025-07-27 at 3 24 35 AM
Screenshot 2025-07-27 at 3 24 50 AM

### DIFF
--- a/packages/grafana-llm-app/src/components/AppConfig/LLMConfig.tsx
+++ b/packages/grafana-llm-app/src/components/AppConfig/LLMConfig.tsx
@@ -82,7 +82,7 @@ export function LLMConfig({
 
   const hasExistingModelConfigurations = () => {
     const modelMapping = settings.models?.mapping;
-    return modelMapping && Object.keys(modelMapping).some(key => modelMapping[key as keyof typeof modelMapping]);
+    return modelMapping && Object.keys(modelMapping).some((key) => modelMapping[key as keyof typeof modelMapping]);
   };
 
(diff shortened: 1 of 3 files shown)


packages/grafana-llm-app/src/components/AppConfig/LLMConfig.tsx
```diff
@@ -234,21 +291,33 @@ export function LLMConfig({
               </Card.Figure>
             </Card>
           </div>
+          </div>
         )}
+        <div className={s.providerSection}>
+          <div className={s.sectionHeader}>
+            <h4>Bring Your Own Service {(llmOption === 'openai' || llmOption === 'anthropic' || llmOption === 'custom') && <Badge text="Currently Active" color="green" />}</h4>
+          </div>
```
sd2k, line 299: this looks a bit unusual when `allowGrafanaManagedLLM` is falsey because it says 'Currently Active' but is the only option. It made me wonder what the alternatives are 😅 I wonder if we could also gate it behind the same variable?

iamrajiv:
sure thanks for review @sd2k i pushed the relevant changes and take a look also ci is failed its not related to my changes seems like component renders should i fix in this pr or?

iamrajiv:
ah my bad @sd2k can you check now?

sd2k (approved):
Nice thanks again @iamrajiv!

PR #766, opened 2025-08-02
chore(health): Improve error field when OpenAI health check fails

### LINKED ISSUE #101: Improve `error` field when OpenAI health check fails
The [`error` field of the OpenAI health check](https://github.com/grafana/grafana-llm-app/blob/27588c9eb4f7f13b6ef2fb48ee80dbe32d3f8961/pkg/plugin/health.go#L30C5-L30C5) is currently a stringified Go error, which is tricky for the frontend to parse. It would be nicer for the frontend if we returned something better, such as the actual response body as an `any` type. This could even appear in a separate nullable field, in case there _is_ no response body (e.g. if the HTTP request failed to send for some reason). The details would then look something like...

```json
{
  "configured": true,
  "ok": false,
  "error": "unexpected status code 401: {<the `response` object below>}",
  "response": {
    "error": {
      "message": "You didn't provide an API key. You need to provide your API key in an Authorization header using Bearer auth (i.e. Authorization: Bearer YOUR_KEY), or as the password field (with blank username) if you're accessing the API from your browser and are prompted for a username and password. You can obtain an API key from https://platform.openai.com/account/api-keys.",
      "type": "invalid_request_error",
      "param": null,
      "code": null
    }
  }
}
```

iamrajiv:
Fixes #101

### DIFF
--- a/packages/grafana-llm-app/pkg/plugin/health.go
+++ b/packages/grafana-llm-app/pkg/plugin/health.go
@@ -3,6 +3,7 @@ package plugin
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 
 	"github.com/grafana/grafana-plugin-sdk-go/backend"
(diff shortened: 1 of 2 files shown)


sd2k (approved):
LGTM! thanks 🙏

iamrajiv:
great thanks for review @sd2k

PR #834, opened 2025-12-11
Chore: update @modelcontextprotocol/sdk

ryantxu:
Avoiding: https://www.cve.org/CVERecord?id=CVE-2025-66414


sd2k:
nope I think you're good, those tests were failing on main, fixed in #835. a rebase would work!

Issue #856, opened 2026-03-04
Health check is passing but is having issues with actual LLM functions


robenie-cwt:
I'm currently having issues with the Grafana LLM App. It is passing health checks 

<img width="1589" height="1038" alt="Image" src="https://github.com/user-attachments/assets/46cff49a-513a-4c61-b426-96f4fdd1f199" />

But when attempting to use LLM features, it always fails. For example, in this case the auto generation of description:

<img width="1220" height="758" alt="Image" src="https://github.com/user-attachments/assets/0056e85b-7be4-455b-b3fe-56595ef514a7" />

Any idea what is causing this and what is the fix?

milon619:
I am facing the same issue, saw this reported [here](https://github.com/grafana/grafana-llm-app/issues/857) as well however I suspect whether it's due to the message size because as per [this PR](https://github.com/grafana/grafana/pull/108414) the default value is set to 8MB which I beleive should be sufficient for most trivial use-cases.

I am on Grafana 12.4.1 and the plugin version is 1.0.6 and using a custom OpenAI compatible provider. Health check passes, however auto-generate fails silently.

<img width="1264" height="330" alt="Image" src="https://github.com/user-attachments/assets/c8510c1d-a49d-4a97-9342-1ae8736e6478" />

milon619:
@jacano1969 I was under the assumption that the default has been raised as per [this](https://github.com/grafana/grafana/pull/108414) PR to 8MB. So isn't that the case? Did you notice anything in the logs indicating the message size limit is indeed the reason that you have described in Problem 2 ?

> The default message size limit is 64KB. This limit is exceeded in two common scenarios:

Also I am not sure whether I followed you correctly on the MCP issue that you described. Currently I don't see any option in the plugin config customising the MCP servers. Could you provide some guidance on how to achieve this? 

> Reducing the number of exposed MCP tools (by switching to an external mcp-grafana instance instead of the built-in one)

jacano1969:
#milon

> [@jacano1969](https://github.com/jacano1969) I was under the assumption that the default has been raised as per [this](https://github.com/grafana/grafana/pull/108414) PR to 8MB. So isn't that the case? Did you notice anything in the logs indicating the message size limit is indeed the reason that you have described in Problem 2 ?
> 
> > The default message size limit is 64KB. This limit is exceeded in two common scenarios:
> 


Sorry, I've been very busy at work and haven't been able to reply sooner.

[...]

PR #882, opened 2026-04-14
feat(llm): make stream timeout configurable and increase default to 60s

gitdoluquita:
Addresses #856. The hardcoded 10s timeout in useLLMStream was too
aggressive for CPU-only inference or models under load. The default
is now 60s and callers can pass a custom timeout value.

### DIFF
--- a/packages/grafana-llm-frontend/src/llm.ts
+++ b/packages/grafana-llm-frontend/src/llm.ts
@@ -605,7 +605,7 @@ export enum StreamStatus {
  * A constant representing the timeout value in milliseconds.
  * @type {number}
  */
-export const TIMEOUT = 10000;
+export const TIMEOUT = 60000;
 
 /**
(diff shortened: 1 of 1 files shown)


packages/grafana-llm-frontend/src/llm.test.ts
```diff
@@ -114,3 +115,9 @@ describe("extractContent", () => {
     });
   });
 });
+
+describe("TIMEOUT", () => {
+  it("should have a default value of 60 seconds", () => {
+    expect(TIMEOUT).toBe(60000);
+  });
+});
```
sd2k: not sure about this test 😂

gitdoluquita: It's definetly testing something... 🤦 
I'll update it!

sd2k (approved):
LGTM

PR #910, opened 2026-05-06
feat: allow custom model strings to avoid unrecognized model error

phoenixiyer:
This PR implements an enhancement to the LLM client by resolving the backend TODOs in `llm_provider.go`. It allows custom, unrecognized model strings to safely pass through parsing instead of erroring, enabling them to map dynamically or safely fall back to default models configured in settings.

### Changes:
1. **Support Unrecognized Model Strings**: Modified `ModelFromString` and `UnmarshalJSON` in `llm_provider.go` to successfully parse non-empty custom model strings as `Model(m)` instead of throwing an error.
2. **Enable Dynamic Fallbacks**: Allows custom default models specified in the settings to be used properly, integrating cleanly with existing mapping fallbacks.
3. **Test Adaptations**: Updated `llm_provider_test.go` to assert that custom model strings are successfully preserved rather than treated as errors.

Signed-off-by: ArunKG <g.arunkumar27@gmail.com>

### DIFF
--- a/packages/grafana-llm-app/pkg/plugin/llm_provider_test.go
+++ b/packages/grafana-llm-app/pkg/plugin/llm_provider_test.go
@@ -29,9 +29,9 @@ func TestModelFromString(t *testing.T) {
 
 		// unknown models
 		{
-			input:    "invalid_model",
-			expected: "",
-			wantErr:  true,
+			input:    "custom_model",
+			expected: Model("custom_model"),
+			wantErr:  false,
 		},
 		{
 			input:    "",
@@ -120,9 +120,9 @@ func TestModelUnmarshalJSON(t *testing.T) {
 
 		// unknown models
 		{
-			input:    []byte(`"invalid_model"`),
-			expected: "",
-			wantErr:  true,
+			input:    []byte(`"custom_model"`),
+			expected: Model("custom_model"),
+			wantErr:  false,
 		},
 		{
 			input:    []byte(`""`),
--- a/packages/grafana-llm-app/pkg/plugin/llm_provider.go
+++ b/packages/grafana-llm-app/pkg/plugin/llm_provider.go
@@ -21,23 +21,26 @@ const (
 	ModelLarge = "large"
 )
 
-// UnmarshalJSON accepts either OpenAI named models for backwards
-// compatability, or the new abstract model names.
 func ModelFromString(m string) (Model, error) {
+	if m == "" {
+		return "", fmt.Errorf("model name cannot be empty")
+	}
 	switch {
 	case m == ModelLarge || (strings.HasPrefix(m, "gpt-4") && !strings.Contains(m, "-mini")):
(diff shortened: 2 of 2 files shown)


packages/grafana-llm-app/pkg/plugin/llm_provider_test.go
```diff
@@ -121,8 +121,8 @@ func TestModelUnmarshalJSON(t *testing.T) {
 		// unknown models
 		{
 			input:    []byte(`"invalid_model"`),
-			expected: "",
-			wantErr:  true,
+			expected: Model("invalid_model"),
+			wantErr:  false,
```
sd2k: it would be good to change the text in here to something like `custom_model` rather than `invalid_model` to reflect the new behaviour.

sd2k (commented):
Thanks for the contribution, this seems reasonable, have you tested a custom build of the plugin with this change in Grafana?

phoenixiyer:
Thank you for the review! 

I have updated the unit test assertions in `llm_provider_test.go` to rename `invalid_model` to `custom_model` to better reflect the new support for custom/unrecognized model configurations.

Regarding testing, I have verified that all Go backend unit tests pass successfully. As this resolves the core parsing constraints for unrecognized models on the backend, it allows custom model settings to pass through cleanly and fall back safely to configured default models in the settings mapping.

AmirF194 (commented):
While this PR is touching both model parsers, it may be worth unifying them here, because they currently disagree and this change carries the disagreement forward.

`ModelFromString` routes any model containing `-mini` to `base` (added in 05191dc), but `Model.UnmarshalJSON` has no `-mini` exclusion, so any `gpt-4*` string routes to `large`. Every incoming chat completion request goes through `UnmarshalJSON` (the stream and HTTP paths), while the Azure deployment mapping uses `ModelFromString` (azure_provider.go), so the same model string lands on different tiers depending on the path. I verified with a small differential test in a clean container (golang:1.26) against both current main (92f56b2) and this branch (7b1479d):

```
"gpt-4o-mini":  ModelFromString="base"  but UnmarshalJSON="large"
"gpt-4.1-mini": ModelFromString="base"  but UnmarshalJSON="large"
```

The user-visible effect is that a chat request for `gpt-4o-mini` is silently served the `large` mapping, the expensive tier.

Since this PR already rewrites both functions, one option is to make `UnmarshalJSON` unmarshal into a plain string and delegate to `ModelFromString`. That removes this divergence permanently, and it would also make the new custom-model passthrough behave identically on both paths (`strings.Trim(dataString, "\"")` can disagree with the string's decoded JSON value in edge cases). Happy to share the test if useful.

Not blocking, and this could equally be a separate change if you'd rather keep this PR focused; the passthrough behavior here looks right to me either way.

For transparency: I used an AI assistant to find and verify this; the outputs above are from actual test runs.

PR #939, opened 2026-06-09
fix(live): set both stream and namespace on plugin Live channels

sd2k:
## Problem

On Grafana **≥ 12.4**, plugin-scoped Grafana Live subscriptions from this plugin (LLM streaming via `streamChatCompletions`, and the MCP `GrafanaLiveTransport`) silently fail. The visible symptom is the **"✨ Auto-generate title/description"** button timing out after 10s with `LLM stream timed out`, and **no `subscribe` frame** for the plugin channel on the WebSocket.

## Root cause

grafana/grafana#117275 (*"Live: Use namespace rather than OrgID"*, in 12.4) renamed `LiveChannelAddress.namespace` → `stream` and changed the frontend to build the channel id from `stream`:

```js
// public/app/features/live/centrifuge/service.ts
const id = `${this.deps.namespace}/${addr.scope}/${addr.stream}/${addr.path}`;
```

This plugin only set the (now-deprecated) `namespace` field, so on 12.4+ `addr.stream` is `undefined` and the channel id becomes:

```
default/plugin/undefined/llm/v1/chat/completions/<uuid>
```

which never routes to the plugin's `SubscribeStream`/`RunStream`. Grafana's `namespace → stream` back-compat shim (`isValidLiveChannelAddress`) runs *after* the id is already built, so it doesn't rescue this path. On Grafana < 12.4 the old code reads `namespace`, so it worked there — matching the reporter's 12.3.3-works / 12.4.3-breaks bisection.

## Fix

Set **both** `stream` and `namespace` (to the same plugin id) via a shared `pluginLiveChannel` helper, so a single plugin build subscribes to the correct channel on both old (`namespace`-reading) and new (`stream`-reading) Grafana. A `LiveChannelAddress & { stream: string }` intersection type keeps this fully type-checked even when building against the pinned pre-12.4 `@grafana/data` (whose `LiveChannelAddress` type has no `stream` field) — no cast needed.

Applied to all three channel construction sites:
- `streamChatCompletions` (`llm.ts`)
- `GrafanaLiveTransport` subscribe/publish channels (`mcp.tsx`)

## Notes

- This is the **plugin-side** fix and works against already-shipped Grafana 12.4.x. The underlying Grafana bug (channel id built before namespace→stream normalization) still affects any plugin using the deprecated `namespace` field and is worth fixing upstream too.
- Added a regression test asserting `streamChatCompletions` subscribes via a channel carrying **both** `stream` and `namespace`.

## Testing

- `tsc --noEmit` — clean
- `eslint` — clean
- `jest` — 28/28 pass (incl. new regression test)

Ref: https://github.com/grafana/grafana/issues/125801

### DIFF
--- a/packages/grafana-llm-frontend/src/constants.ts
+++ b/packages/grafana-llm-frontend/src/constants.ts
@@ -1,9 +1,32 @@
+import { LiveChannelAddress, LiveChannelScope } from "@grafana/data";
 import { logWarning } from "@grafana/runtime";
 
 import { SemVer } from "semver";
 
 export const LLM_PLUGIN_ID = "grafana-llm-app";
 export const LLM_PLUGIN_ROUTE = `/api/plugins/${LLM_PLUGIN_ID}`;
+
+// Grafana 12.4 renamed `LiveChannelAddress.namespace` to `stream` and now
+// builds the channel id from `stream`, ignoring `namespace`. Older versions do
+// the reverse. We must set both so a single plugin build subscribes to the
+// correct channel regardless of which Grafana version is running; omitting
+// `stream` on 12.4+ yields a `plugin/undefined/...` channel that never routes.
+// The intersection adds `stream` to the type without losing type checking on
+// the other fields, even when building against a pre-12.4 `@grafana/data`.
+type PluginLiveChannelAddress = LiveChannelAddress & { stream: string };
+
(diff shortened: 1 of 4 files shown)


nrbrd (approved):
lgtm

PR #940, opened 2026-06-12
chore(version): bump version to v1.0.9

sd2k:
## What

Bumps the plugin and `@grafana/llm` package versions from `1.0.8` → `1.0.9` (patch) and regenerates the changelog.

## Why

The `Version bump and update changelog` workflow now fails at the `git push origin main` step:

```
remote: Permission to grafana/grafana-llm-app.git denied to grafana-plugins-platform-bot[bot].
fatal: ... error: 403
```

Branch protection no longer allows the bot to push directly to `main`. This PR performs the same bump manually so it can land through the normal PR flow.

## Changes (mirrors the workflow steps)

- `npm version patch` in `packages/grafana-llm-app` and `packages/grafana-llm-frontend`
- Regenerated root `package-lock.json`
- Regenerated `CHANGELOG.md` (`generate-changelog -t v1.0.8...HEAD`)

## Follow-up

The `v1.0.9` git tag still needs to be created/pushed after merge (the workflow previously did this after the direct push). Consider updating `version-bump.yml` to open a PR instead of pushing to `main`, and to handle tagging post-merge.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19791,7 +19791,7 @@
     },
     "packages/grafana-llm-app": {
       "name": "@grafana/llm-app",
-      "version": "1.0.8",
+      "version": "1.0.9",
       "license": "Apache-2.0",
       "dependencies": {
         "@emotion/css": "^11.13.5",
@@ -20233,7 +20233,7 @@
     },
     "packages/grafana-llm-frontend": {
       "name": "@grafana/llm",
-      "version": "1.0.8",
+      "version": "1.0.9",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.26.0",
--- a/packages/grafana-llm-app/CHANGELOG.md
+++ b/packages/grafana-llm-app/CHANGELOG.md
@@ -1,5 +1,45 @@
 # Changelog
 
+#### 1.0.9
+
+##### Chores
+
+* **deps:**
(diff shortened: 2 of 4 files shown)


nrbrd (approved):

rvdh:
@sd2k @nrbrd can v1.0.9 be published to the Grafana catalogue please? Would love to test it.

sd2k:
Hey @rvdh, I don't expect v1.0.9 of the plugin to actually make any difference, the changes were to the `@grafana/llm` package used by clients of the LLM app (notably, Grafana) which will need to be updated to v1.0.9 to make use of the fixes.

rvdh:
@sd2k thanks for the quick response! I don't really understand though, as the [PR](https://github.com/grafana/grafana-llm-app/pull/939) states: 

> This is the plugin-side fix and works against already-shipped Grafana 12.4.x. The underlying Grafana bug (channel id built before namespace→stream normalization) still affects any plugin using the deprecated namespace field and is worth fixing upstream too.

But you are saying a new 12.4.x should be made with the new package?

sd2k:
> still affects any plugin using the deprecated namespace field

This should really read:

> still affects any plugin or Grafana version using the deprecated namespace field

So we need to bump `@grafana/llm` in Grafana and backport it to 12.4.x, yeah. I'll get on that today!

sd2k:
https://github.com/grafana/grafana/pull/126530 bumps in Grafana.

PR #941, opened 2026-06-12
ci: open a PR for version bumps instead of pushing to main

sd2k:
## What

Reworks the release-tagging flow so it no longer relies on the bot pushing directly to `main`:

- **`version-bump.yml`** — instead of committing the bump and tag straight to `main`, it now creates a `version-bump/<version>` branch, pushes it, and opens a PR (using the same plugins-platform-bot app token).
- **`tag-release.yml`** (new) — triggers on pushes to `main` that touch `packages/grafana-llm-app/package.json`, reads the version, and creates/pushes the `vX.Y.Z` tag if it doesn't already exist.

## Why

The `Version bump and update changelog` workflow currently fails at `git push origin main`:

```
remote: Permission to grafana/grafana-llm-app.git denied to grafana-plugins-platform-bot[bot].
fatal: ... error: 403
```

Branch protection no longer allows direct pushes to `main`. (#940 is the manual `v1.0.9` bump done while this was broken; this PR fixes the automation going forward.)

## How tagging works now

The git tag isn't a trigger for any workflow in this repo (publish/CD and npm-release are all `workflow_dispatch`; `push.yml` runs on push to `main`), so it's purely a release marker / `git describe` anchor. Moving it to a post-merge step is safe:

- When a version-bump PR merges, the bump commit lands on `main` → `tag-release.yml` runs → tag created.
- The step is **idempotent**: it skips if the tag already exists, so renovate/dependency PRs that touch `package.json` (without changing the `version` field) are harmless no-ops.

## Notes / assumptions

- Assumes the `plugins-platform-bot-app` GitHub App has **pull-requests: write** (needed for `gh pr create`) and that it *can* push non-`main` branches and tags (the 403 was specific to the protected `main` branch).
- `tag-release.yml` reuses the exact Vault + app-token pattern already in `version-bump.yml`, so zizmor findings are identical to what's already accepted on `main` (no new finding types). The version-bump edits also reduced a few `template-injection` infos by switching to env vars.

### DIFF
--- a/.github/workflows/tag-release.yml
+++ b/.github/workflows/tag-release.yml
@@ -1,0 +1,73 @@
+name: Tag release
+
+# When a version bump lands on `main` (via the PR opened by `version-bump.yml`),
+# create and push the matching `vX.Y.Z` git tag. This replaces the inline tagging
+# that `version-bump.yml` used to do before direct pushes to `main` were blocked
+# by branch protection.
+#
+# The job is idempotent: it only creates the tag if it doesn't already exist, so
+# unrelated pushes that touch `package.json` (e.g. dependency bumps) are no-ops.
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - packages/grafana-llm-app/package.json
+
+permissions:
+  contents: read
+  id-token: write
+
+jobs:
+  tag:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Get secrets from Vault
+        id: get-secrets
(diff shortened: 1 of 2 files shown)


sd2k:
Superseded by #942, which migrates the version-bump workflow to the GitHub App Token Broker (mirroring grafana-assistant-app#6781). That restores direct push to `main`, so the PR-based-bump + tag-release workaround here is no longer needed.

PR #947, opened 2026-06-16
chore: make grafanaDependency prerelease-inclusive

tolzhabayev:
grafana-com PR #17309 made the plugins publish API reject Grafana-owned plugins whose `grafanaDependency` lower bound isn't prerelease-inclusive, so the next publish of this plugin would fail.

This updates the range from `>=9.5.2` to `>=9.5.2-0` so prerelease Grafana builds are matched and publishing keeps working. No behavior change for released Grafana versions.

Ref: https://github.com/grafana/grafana-com/pull/17309

### DIFF
--- a/packages/grafana-llm-app/src/plugin.json
+++ b/packages/grafana-llm-app/src/plugin.json
@@ -28,7 +28,7 @@
   },
   "includes": [],
   "dependencies": {
-    "grafanaDependency": ">=9.5.2",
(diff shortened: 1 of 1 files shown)


sd2k:
@tolzhabayev could you rebase this on main? CI should pass then and I'll get this in

tolzhabayev:
@sd2k done

sd2k (approved):

PR #948, opened 2026-06-16
fix(e2e): bump Playwright Docker image to v1.61.0-noble

sd2k:
## What

Bumps the Playwright Docker image in `packages/grafana-llm-app/docker-compose.yaml` from `v1.60.0-noble` to `v1.61.0-noble` (both `playwright-server` and `playwright-runner` services).

## Why

The Dockerized Playwright services run `npm install --legacy-peer-deps` at container start, which resolves `@playwright/test` (`^1.52.0`) to the latest 1.x — currently **1.61.0**. The image, however, was pinned to `v1.60.0-noble`, whose bundled browsers don't match. Playwright refuses to run on a version mismatch:

```
Error: browserType.launch: Executable doesn't exist at /ms-playwright/chromium_headless_shell-1228/chrome-headless-shell-linux64/chrome-headless-shell
╔══════════════════════════════════════════════════════╗
║ -  current: mcr.microsoft.com/playwright:v1.60.0-noble ║
║ - required: mcr.microsoft.com/playwright:v1.61.0-noble ║
╚══════════════════════════════════════════════════════╝
```

This was failing the `tests` (Dockerized Playwright E2E) check on `main`, independent of any individual PR's changes (e.g. #947).

Follows the established manual-bump pattern (0b88237b, 3cba9c3a, 2629775a).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #949, opened 2026-06-16
fix(ci): scope create-github-app-token in version-bump workflow

sd2k:
## What

Scopes the `actions/create-github-app-token` step in `.github/workflows/version-bump.yml` to:
- `repositories: ${{ github.event.repository.name }}` — this repo only
- `permission-contents: write` — the only permission the workflow uses

## Why

zizmor's `github-app` audit (now run at v1.25.2, **high** severity, which fails the build) flags two issues with this step, introduced when the action was bumped to v3.2.0 in #927:

```
error[github-app]: dangerous use of GitHub App tokens
  --> ./.github/workflows/version-bump.yml:51:11
   | owner: ${{ github.repository_owner }}
   | token granted access to all repositories for this owner's app installation
   = tip: use `repositories: 'repo1,repo2'` to scope the token to specific repositories

error[github-app]: dangerous use of GitHub App tokens
  --> ./.github/workflows/version-bump.yml:47:15
   | uses: actions/create-github-app-token@... # v3.2.0
   | app token inherits blanket installation permissions
   = tip: specify at least one `permission-<name>` input to limit the token's permissions
```

The workflow only uses the token to check out and `git push` the version-bump commit and tags, so `contents: write` scoped to this single repository is sufficient. This was failing the `zizmor` check on `main`, independent of any individual PR (e.g. #947).

### DIFF
--- a/.github/workflows/version-bump.yml
+++ b/.github/workflows/version-bump.yml
@@ -49,6 +49,11 @@ jobs:
           app-id: ${{ fromJSON(steps.get-secrets.outputs.secrets).GITHUB_APP_ID }}
           private-key: ${{ fromJSON(steps.get-secrets.outputs.secrets).GITHUB_APP_PRIVATE_KEY }}
           owner: ${{ github.repository_owner }}
+          # Scope the token to this repository and to the only permission this
+          # workflow needs (pushing the version-bump commit and tags), rather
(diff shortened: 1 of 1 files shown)


sd2k:
Superseded by #950, which combines this with the Playwright image fix (#948) — both are required checks, so they need to land together to turn CI green.

PR #953, opened 2026-07-08
fix: return after writing error responses in resource handlers

AmirF194:
## Problem

Four error branches in `resources.go` write an error response via `handleError` and then fall through instead of returning: the `createProvider` error and `errBadRequest` branches in both `handleModels` and `handleChatCompletions`. The handler then continues and writes a second JSON document into the same response body.

The easiest way to hit this is the Azure provider with a model that has no deployment mapping. The client receives:

```
{"error":"bad request: no deployment found for model: base"}{"id":"","object":"","created":0,...}
```

which is the 400 error followed by a marshaled zero-value completion, so anything parsing the body as JSON fails. With no provider configured, both handlers return two concatenated error documents instead.

## Fix

Add the missing `return` after each of the four `handleError` calls. No behavior change on success paths; error responses are now exactly one JSON document.

## Verification

Ran in a clean golang:1.26 container: the new regression test `TestErrorResponsesAreSingleJSONDocument` (three cases, driven through `CallResource` like the existing `TestCallResource`) fails on current main with the corrupted bodies shown above, and passes with this change. The full `./pkg/...` suite passes and `go vet` is clean.

For transparency: I used an AI assistant to find, fix, and verify this; the outputs above are from actual test runs.

### DIFF
--- a/packages/grafana-llm-app/pkg/plugin/resources.go
+++ b/packages/grafana-llm-app/pkg/plugin/resources.go
@@ -438,6 +438,7 @@ func (a *App) handleModels() http.HandlerFunc {
 		if err != nil {
 			log.DefaultLogger.Error("LLM provider has invalid configuration", "err", err)
 			handleError(w, errors.New("LLM provider has invalid configuration"), http.StatusUnprocessableEntity)
+			return
 		}
 		if llmProvider == nil {
 			handleError(w, errors.New("must configure an LLM provider"), http.StatusUnprocessableEntity)
@@ -450,6 +451,7 @@ func (a *App) handleModels() http.HandlerFunc {
 		models, err := llmProvider.Models(r.Context())
 		if errors.Is(err, errBadRequest) {
 			handleError(w, err, http.StatusBadRequest)
+			return
 		} else if err != nil {
 			handleError(w, err, http.StatusInternalServerError)
 			return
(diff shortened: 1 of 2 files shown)


AmirF194:
No rush, but flagging this in case it slipped through since it's been about a week. Still clean and mergeable, happy to rebase or add whatever would help review.

gerboland (approved):
Sorry for the delay, you're quite right that's a necessary fix. Thank you

AmirF194:
No worries on the timing, thanks for taking a look. Let me know if there's anything else you'd like changed before merging.

gerboland:
Hi Amir, we have a new policy that requires all commits to be signed before we can land them:
https://community.grafana.com/t/action-required-signed-commits-mandatory-for-all-grafana-repositories/163404
I cannot merge until that is done. Apologies for the additional effort.

AmirF194:
Good catch, done. Re-signed and force-pushed the commit (SSH signing); the signed-commits check is green now.

AmirF194:
The only thing left is on your side: the workflow runs are sitting at "action required" because this comes from a fork, so they need your approval before the merge can go through.

Also a heads up, I pushed the branch again a few minutes ago while double checking the signature. Same content, the tree hash is unchanged, head is now cd640a5 and still shows verified. Sorry for the extra noise.

gerboland:
Perfect, thank you!

AmirF194:
Thanks for merging, and for flagging the signed-commits requirement, glad it was a quick fix on my end.

PR #990, opened 2026-08-11
fix(llmclient): bound health-check response read with io.LimitReader

### LINKED ISSUE #307: nit: limit the amount of reading
just in case, we should consider an [`io.LimitReader`](https://pkg.go.dev/io#LimitReader) here. Otherwise it could hang here for ages/ever.

https://github.com/grafana/grafana-llm-app/blob/llmclient/v0.1.2/llmclient/llmclient.go#L116C9-L116C36

AmirF194:
## What this does

`llmProvider.Enabled()` reads the app's own health-check response with a plain `io.ReadAll(resp.Body)`, with no size limit. Wraps it in `io.LimitReader(resp.Body, 1024*1024)`, matching the limit already used for the same purpose in `vectorapi.go`, `embed/openai.go`, and `resources.go`.

Fixes #307, filed by @matryer: "just in case, we should consider an `io.LimitReader` here. Otherwise it could hang here for ages/ever."

## Testing

- New regression test (`TestEnabled_LimitsResponseBodyRead`) wraps the HTTP transport with a byte counter and serves a 5MiB non-JSON health response: fails on main (reads all 5MiB), passes on this branch (read capped near 1MiB).
- Full `llmclient` package suite: `go test ./...` green (Go 1.26, matching `go.mod`).
- `gofmt -l` and `go vet ./...` clean.
- `llmclient/` has no CI workflow of its own in this repo (only `packages/grafana-llm-app` is wired into `run-tests.yml`/`is-compatible.yml`/`push.yml`), so the package's own test suite above is the only gate that applies to this change.

### DIFF
--- a/llmclient/llmclient.go
+++ b/llmclient/llmclient.go
@@ -129,7 +129,7 @@ func (o *llmProvider) Enabled(ctx context.Context) (bool, error) {
 	if resp.StatusCode != http.StatusOK {
 		return false, nil
 	}
-	body, err := io.ReadAll(resp.Body)
+	body, err := io.ReadAll(io.LimitReader(resp.Body, 1024*1024))
 	if err != nil {
 		return false, fmt.Errorf("read response body: %w", err)
 	}
--- a/llmclient/llmclient_test.go
+++ b/llmclient/llmclient_test.go
@@ -1,12 +1,14 @@
 package llmclient
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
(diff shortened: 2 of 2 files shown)


AmirF194:
No rush, just checking in. This is ready for review whenever you have time, happy to rebase or split it if that helps.

sd2k (approved):
Cheers!

AmirF194:
Thanks for the quick review and merge!

PR #993, opened 2026-08-25
docs: explain unconfigured LLM provider states

### LINKED ISSUE #363: Update LLM frontend docs to cover unconfigured status
[Our readme here](https://github.com/grafana/grafana-llm-app/tree/main/packages/grafana-llm-frontend) doesn't yet cover how to handle the different LLM app config options, including the new "unconfigured" status added in https://github.com/grafana/grafana-llm-app/pull/350 .

The general UX approach we'd like to model is:
* activate LLM features if they're enabled (either Grafana-provided or other config)
* hide LLM features from the user if they are disabled / opted-out
* prompt the user that an admin can opt in to LLM-powered features in the LLM plugin if status is unconfigured.

Once our readme has a trivial example, I can start nudging different squads to use this approach (plus get some UX input on how they should do it).

loglapa:
## What this changes

- document how `llm.health()` distinguishes configured, unconfigured, and unhealthy providers
- clarify that an unconfigured provider is a normal setup state rather than a backend outage
- include a compact branching example for plugin authors

## Validation

- `npm run format:check -- --filter=@grafana/llm`
- `git diff --check`

Validation was run in Docker.

Fixes #363

### DIFF
--- a/packages/grafana-llm-frontend/README.md
+++ b/packages/grafana-llm-frontend/README.md
@@ -55,6 +55,26 @@ global.TransformStream = TransformStream;
 
 ## Usage
 
+### Handle provider configuration states
+
+Before rendering LLM-powered features, use `llm.health()` to distinguish between an available provider, an unconfigured provider, and a provider that an administrator has configured but that is currently unavailable:
+
+```typescript
+import { llm } from '@grafana/llm';
+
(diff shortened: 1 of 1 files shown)


sd2k (approved):
Thank you <3

PR #995, opened 2026-08-28
feat: support configurable Azure API versions

### LINKED ISSUE #858: Support configurable Azure OpenAI API version
## Summary

Currently, the grafana-llm-app plugin does not allow configuration of the Azure OpenAI API version. The API version is hardcoded in the underlying `go-openai` library, which may not be compatible with all Azure OpenAI deployments or may require updating when Azure releases new API versions.

## Problem

When deploying Azure OpenAI models through this plugin, users cannot specify which API version to use (e.g., `2024-08-01-preview`, `2024-10-01-preview`, etc.). This can lead to:

- Incompatibility with specific Azure OpenAI deployments
- Inability to use newly released API features
- Forced reliance on whatever default version is used by the underlying `go-openai` library

## Proposed Solution

Add a configurable `apiVersion` field to the Azure OpenAI settings, similar to how other settings like `url` and `azureModelMapping` are already configurable.

### Example Configuration

```yaml
provider: azure
openAI:
  url: https://<resource>.openai.azure.com
  apiVersion: "2024-08-01-preview"  # New configurable field
  azureModelMapping:
    - ["base", "gpt-35-turbo"]
    - ["large", "gpt-4-turbo"]

loglapa:
## Summary

- add an optional `openAI.apiVersion` setting for Azure OpenAI
- expose the setting in the Azure configuration UI and provisioning documentation
- apply it to `go-openai`'s Azure client only when non-empty, preserving the SDK default otherwise
- cover UI visibility/editing, settings JSON round-trip, and the actual `api-version` query parameter for streaming and non-streaming requests

## Why

Azure OpenAI deployments can require a different API version from the one currently hardcoded by the client library. The plugin had no way to select that version, so otherwise valid deployments could not be used without changing the dependency.

The setting is intentionally optional and free-form. Azure version identifiers evolve (including dated and preview forms), while the SDK safely URL-encodes the query value. Leaving the field blank retains the exact existing behavior.

## Verification

- frontend Jest: 3/3 suites, 5 tests passed
- frontend TypeScript, ESLint, and Prettier checks passed
- focused Azure transport/settings tests under `go test -race` passed
- `go test ./... -count=1` passed
- `go vet ./...` passed
- `golangci-lint run ./...`: 0 issues
- `gofmt` and `git diff --check`

Closes #858

### DIFF
--- a/packages/grafana-llm-app/README.md
+++ b/packages/grafana-llm-app/README.md
@@ -70,6 +70,7 @@ apps:
       provider: azure
       openAI:
         url: https://<resource>.openai.azure.com
+        apiVersion: "2024-10-21"
         azureModelMapping:
           - ["base", "gpt-35-turbo"]
           - ["large", "gpt-4-turbo"]
@@ -80,6 +81,7 @@ apps:
 where:
 
 - `<resource>` is your Azure OpenAI resource name
+- `apiVersion` is optional; omit it to use the `go-openai` SDK default
 - the `azureModelMapping` field contains `[model, deployment]` pairs so that features know
   which Azure deployment to use in place of each model you wish to be used.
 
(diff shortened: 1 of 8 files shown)


gerboland (approved):
Sorry for the delay, looks good, thank you

PR #1001, opened 2026-09-17
fix: send JSON when saving plugin settings to gcom

gerboland:
## Summary
- The provisioned-plugins update endpoint (`/instances/{id}/provisioned-plugins/{id}`) moved to a Go service that only accepts `application/json`, so saving LLM app settings now fails with `HTTP error 415` from gcom.
- Switch the save-settings request to send JSON instead of `application/x-www-form-urlencoded`. The old Node.js endpoint accepts JSON too, so this is compatible either way.

## Test plan
- `go test ./pkg/plugin/...` passes (updated `TestMergeSecureJSONData` for the new JSON body shape).
- `go build ./...` and `go vet ./...` clean.


sd2k (approved):
yep lgtm, similar to our other plugins

